PR #115, opened 2024-07-03
Use cargo nextest to speed up the testing workflow

Hermann-Core:
This pull request adjusted the CI pipeline to incorporate `cargo nextest run` instead of `cargo test` for running tests. This change is meant to significantly improve the speed of the testing process.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
     - name: Cache dependencies
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: |
           ~/.cargo/bin
@@ -40,3 +40,4 @@ jobs:
       run: |
         cargo build --workspace --all-features
         cargo nextest run --workspace --all-features
+        
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,12 +1,6 @@
 name: Rust CI
 
-on:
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -22,10 +22,13 @@ jobs:
       with:
         toolchain: stable
 
+    - name: Install Nextest test runner
+      uses: taiki-e/install-action@nextest
```
IngridPuppet, line 26: This move will certainly be impactful to reduce pipeline times. 

In addition, can you set up the caching of dependencies? It is generally more significant toward this aim. Check this link to see what I mean: https://stackoverflow.com/questions/64634535/github-actions-cache-rust-artifacts

IngridPuppet (changes requested):

.github/workflows/rust.yml
Hermann-Core: I was looking for a way to cache the artifacts to avoid rebuilding everything each time. This will further reduce the build time. Thank you for the link.

Hermann-Core: The combination of **nextest** and **artifact caching** have significantly reduced the build and test time from over 4 minutes to just 43 seconds. Check it out.

Screenshot from 2024-07-04 11-33-57

IngridPuppet: Great 🎉

Can you further update the syntax of the workflow and bump up some action versions to avoid the following warnings?
 
image

Hermann-Core: thanks for addressing that. Fixed now

IngridPuppet (approved):

PR #116, opened 2024-07-08
76 change the signature of the mount plugin method to accept a mutable reference

ndefokou:
@IngridPuppet @Christiantyemele Please review this

### DIFF
--- a/generic-server/src/plugin/container.rs
+++ b/generic-server/src/plugin/container.rs
@@ -1,4 +1,4 @@
-use std::sync::Mutex;
+use std::sync::{Mutex,Arc};
 use std::collections::{HashMap, HashSet};
 
 use axum::Router;
@@ -12,34 +12,35 @@ pub enum PluginContainerError {
     Unloaded,
     PluginErrorMap(HashMap<String, PluginError>),
 }
-
-pub struct PluginContainer<'a> {
+pub struct PluginContainer {
     loaded: bool,
     collected_routes: Vec<Router>,
-    plugins: &'a Mutex<Vec<Box<dyn Plugin + Send>>>,
+    plugins: Arc<Vec<Arc<Mutex<dyn Plugin + Send>>>>,
 }
 
-impl<'a> Default for PluginContainer<'a> {
+impl<'a> Default for PluginContainer {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<'a> PluginContainer<'a> {
+impl<'a> PluginContainer {
     /// Instantiate an object aware of all statically registered plugins
     pub fn new() -> Self {
         Self {
             loaded: false,
             collected_routes: vec![],
-            plugins: &*PLUGINS,
+            plugins: Arc::clone(&PLUGINS),
         }
     }
 
     /// Search loaded plugin based on name string
-    pub fn find_plugin(&self, name: &str) -> Option<usize> {
-        let plugins = self.plugins.lock().unwrap();
-        plugins.iter().position(|plugin| name == plugin.name())
-    }    
+    pub fn find_plugin(&self, name: &str) -> Option<Arc<Mutex<dyn Plugin + Send>>> {
+        self.plugins.iter().find_map(|arc_plugin| {
+            let plugin = arc_plugin.lock().unwrap();
+            (plugin.name() == name).then_some(Arc::clone(&arc_plugin))
+        })
+    }   
     
     /// Load referenced plugins
     ///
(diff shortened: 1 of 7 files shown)


generic-server/src/plugin/container.rs
```diff
@@ -185,87 +192,143 @@ mod tests {
...
+        // Verify find_plugin method
         assert!(container.find_plugin("first").is_some());
         assert!(container.find_plugin("second").is_some());
         assert!(container.find_plugin("non-existent").is_none());
 
-        // The actual routes collected are actually hard to test
-        // given that axum::Router seems not to provide public
-        // directives to inquire internal state.
-        // See: https://github.com/tokio-rs/axum/discussions/860
+        // Verify collected routes
```
IngridPuppet, line 217: Do not remove the extensive comment that was there.
```suggestion
        // Verify collected routes
        // The actual routes collected are actually hard to test
        // given that axum::Router seems not to provide public
        // directives to inquire internal state.
        // See: https://github.com/tokio-rs/axum/discussions/860
```

generic-server/src/plugin/container.rs
```diff
@@ -185,87 +192,143 @@ mod tests {
...
+        assert!(container.load().is_ok()); // Load again, should succeed without errors
+    
+        // Verify collected routes
         assert_eq!(container.collected_routes.len(), 2);
     }
-
+    
     #[test]
     fn test_loading_with_duplicates() {
+        // Mock plugins for testing with duplicate name "second"
```
IngridPuppet, line 246: Make sure your changes do not alter the narrative of this test.

generic-server/src/plugin/container.rs
```diff
@@ -185,87 +192,143 @@ mod tests {
...
         assert_eq!(container.collected_routes.len(), 1);
     }
 
+
     #[test]
     fn test_route_extraction_without_loading() {
+        // Mock plugins for testing
+        let plugins: Arc<Vec<Arc<Mutex<dyn Plugin>>>> = Arc::new(vec![
+            Arc::new(Mutex::new(FirstPlugin {})),
+            Arc::new(Mutex::new(FaultyPlugin {})),
```
IngridPuppet, line 318: Beware, you are relying too much on ChatGPT.
```suggestion
            Arc::new(Mutex::new(SecondPlugin {})),
```

generic-server/src/plugin/mod.rs
```diff
@@ -1,18 +1,20 @@
...
         #[cfg(feature = "plugin-index")]
-        Box::<index::IndexPlugin>::default(),
+        Arc::new(Mutex::new(index::IndexPlugin::default())),
         #[cfg(feature = "plugin-did_endpoint")]
-        Box::<did_endpoint::plugin::DidEndpointPlugin>::default(),
+        Arc::new(Mutex::new(did_endpoint::plugin::DidEndpointPlugin::default())),
         #[cfg(feature = "plugin-oob_messages")]
-        Box::<oob_messages::plugin::OOBMessagesPlugin>::default(),
-    ];
+        Arc::new(Mutex::new(oob_messages::plugin::OOBMessagesPlugin::default())),
```
IngridPuppet, line 18: I suggest this more concise syntax. Also, format the file.
```rust
        #[cfg(feature = "plugin-index")]
        Arc::<Mutex::<index::IndexPlugin>>::default(),
        #[cfg(feature = "plugin-did_endpoint")]
        Arc::<Mutex::<did_endpoint::plugin::DidEndpointPlugin>>::default(),
        #[cfg(feature = "plugin-oob_messages")]
        Arc::<Mutex::<oob_messages::plugin::OOBMessagesPlugin>>::default(),
```

IngridPuppet (changes requested):
Great PR!

generic-server/src/plugin/container.rs
IngridPuppet: Check the original test. It asserts that the logic raises a DuplicateEntry error. Does your changes do that?
image

ndefokou: yes sir when you try to  load plugins with duplicate names

IngridPuppet: An assert is different than printing messages. Use an assert on the result, instead of the match.

```rust
assert_eq!(result.unwrap_err(), PluginContainerError::DuplicateEntry);
```

ndefokou: okay sir thanks i have correct

IngridPuppet (approved):

PR #119, opened 2024-07-10
Changing the signature of the mount plugin method to accept a mutable reference

ndefokou:


IngridPuppet:
Please merge locally and push. The other PR will then be updated.

PR #131, opened 2024-07-16
refactoring .env loading

ndefokou:


generic-server/src/main.rs
```diff
@@ -4,6 +4,13 @@ use std::net::SocketAddr;
 
 #[tokio::main]
 async fn main() {
+
+    // Load dotenv-flow variables
+    dotenv_flow::from_filename(".env.example").ok();
```
IngridPuppet, line 9: The code must not depend on `.env.example`. Each one will have to create a `.env` file on its computer; it is fine. Remember to add `.env` to `.gitignore`.
```suggestion
    dotenv_flow::dotenv_flow().ok();
```

IngridPuppet (changes requested):

generic-server/src/main.rs
IngridPuppet: Hello

IngridPuppet: There are also conflicts to resolve.

ndefokou: Okay i am just from correcting it

IngridPuppet: I still see `.env.example` in the code.

IngridPuppet:
You shouldn't have removed `.env.example` on functions annotated with `#[cfg(test)]`. It is okay to rely on the file for tests.

IngridPuppet (approved):
Working.

PR #138, opened 2024-07-26
Enhance `did:peer:4` implementation

Hermann-Core:


did-utils/src/methods/did_peer/method.rs
```diff
@@ -473,25 +474,36 @@ impl DidPeer {
     ///
     /// See https://identity.foundation/peer-did-method-spec/#resolving-a-did
     pub fn expand_did_peer_4(&self, did: &str) -> Result<DIDDocument, DIDPeerMethodError> {
-        // Ensure long format by computing did:peer:4 short form alias
-
+        // shorten the did
```
IngridPuppet, line 477: You've removed an important information by changing this comment. The next line actually validates that the argument is a valid long did:peer:4 address, and that should be mentioned.

did-utils/src/methods/did_peer/method.rs
```diff
@@ -473,25 +474,36 @@ impl DidPeer {
...
+        let (base, decoded_hash) = multibase::decode(extracted_hash)
+            .map_err(|_| DIDPeerMethodError::DIDParseError)?;
 
-        let (base, decoded_bytes) = multibase::decode(encoded_document).map_err(|_| DIDPeerMethodError::DIDParseError)?;
-        if base != Base58Btc || decoded_bytes.len() < 2 || decoded_bytes[..2] != MULTICODEC_JSON {
+        // Verify hash
+        let computed_hash = sha256_hash(encoded_document.as_bytes());
+        if Base58Btc != base || decoded_hash.len() < 34 || &decoded_hash[2..] != computed_hash {
             return Err(DIDPeerMethodError::MalformedLongPeerDID);
         }
```
IngridPuppet, line 494: Here, you're checking that the hash in the `did:peer` address is valid. This is already covered by the call to `shorten_did_peer_4`; you're repeating yourself.

did-utils/src/methods/did_peer/method.rs
```diff
@@ -473,25 +474,36 @@ impl DidPeer {
...
+        let (base, decoded_bytes) = multibase::decode(encoded_document)
+            .map_err(|_| DIDPeerMethodError::DIDParseError)?;
 
-        let mut diddoc: DIDDocument = serde_json::from_slice(&decoded_bytes[2..])?;
+        if Base58Btc != base || decoded_bytes.len() < 2 || &decoded_bytes[..2] != MULTICODEC_JSON {
+            return Err(DIDPeerMethodError::MalformedLongPeerDID);
+        }
+        // Deserialize the document
+        let mut diddoc: DIDDocument = serde_json::from_slice(&decoded_bytes[2..])
+            .map_err(|err| DIDPeerMethodError::SerdeError(err))?;
```
IngridPuppet, line 504: This is a format change, and it is creating noise like I told you in an earlier PR. Please reformat the file, we'll need to have it checked in the pipeline.

did-utils/src/methods/did_peer/util.rs
```diff
@@ -84,13 +83,71 @@ fn abbrv_service(value: &mut Value, abbrv_key: &dyn Fn(&str) -> String, abbrv_va
...
+        && diddoc.capability_delegation.is_none()
+        && diddoc.capability_invocation.is_none()
+        && diddoc.service.is_none()
+    {
+        return Err(DIDPeerMethodError::InvalidStoredVariant);
+    }
+
+    // They must not be an id field at the root
+    // All id field within the document must be relative
+    // All references to ressources must be relative
```
IngridPuppet, line 101:
```suggestion
    // There must not be an id field at the root
    // All id fields within the document must be relative
    // All references to resources must be relative
```

did-utils/src/methods/did_peer/util.rs
```diff
@@ -84,13 +83,71 @@ fn abbrv_service(value: &mut Value, abbrv_key: &dyn Fn(&str) -> String, abbrv_va
...
+    if !diddoc.id.is_empty()
+        || !are_all_ids_and_references_relative(diddoc) {
+        return Err(DIDPeerMethodError::InvalidStoredVariant);
+    }
+    Ok(())
+}
```
IngridPuppet, line 107: You seem to disregard the last check in the specs. Can you comment in the code why?
> For verification methods, the controller MUST be omitted if the controller is the document owner. If it is controlled by a DID other than the owner of the document, it MUST be included.

did-utils/src/methods/did_peer/util.rs
```diff
@@ -84,13 +83,71 @@ fn abbrv_service(value: &mut Value, abbrv_key: &dyn Fn(&str) -> String, abbrv_va
     }
 }
 
+pub(super) fn validate_input_document(diddoc: &DIDDocument) -> Result<(), DIDPeerMethodError> {
+    // The document must not be empty
```
IngridPuppet, line 87: Can you source this rule?

did-utils/src/methods/did_peer/util.rs
IngridPuppet, line 1: Auto format the file.

IngridPuppet (changes requested):

did-utils/src/crypto/sha256_hash.rs
```diff
@@ -41,7 +41,7 @@ pub fn sha256_multihash(bytes: &[u8]) -> String {
     ]
     .concat();
 
-    Base58Btc.encode(pack)
+    multibase::encode(Base58Btc, pack)
 }
```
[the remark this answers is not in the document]
Hermann-Core: Already addressed in the doc ticket

did-utils/src/methods/did_peer/method.rs
Hermann-Core: You're right. I haven't correctly checked the code of that method. Let me fix it.

did-utils/src/methods/did_peer/util.rs
Hermann-Core: This is a mistake. Will add the code for that tomorrow.

did-utils/src/methods/did_peer/util.rs
Hermann-Core: It is obvious that creating a `did:peer:4` with an empty document is useless and must be avoid. Aren't you ok with that ?

did-utils/src/crypto/sha256_hash.rs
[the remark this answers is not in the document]
IngridPuppet: You mean on `doc-107-improve-code-documentation-for-the-project`? No it is not addressed. Having comments not in sync with code updates is the one of the most common errors that could persist in a codebase. Kindly address this here, where it should.
https://github.com/adorsys/didcomm-mediator-rs/blob/0f8e07ca0734bfd6a21d0b850149a569fc5bf674/did-utils/src/crypto/sha256_hash.rs#L25-L47

did-utils/src/methods/did_peer/util.rs
IngridPuppet: I had a similar reasoning, that's why I recommend to add a comment explaining that in the code.

However, it is possible the specs have another reasoning, otherwise I don't see the point of adding that rule.

did-utils/src/methods/did_peer/util.rs
IngridPuppet: The wording makes it sound like it is normative.

Please can you update the comment from "// The document must not be empty" to "// The document **should** not be empty"?

Besides, even though it might seem useless, I personally don't think it must be avoided.

did-utils/src/methods/did_peer/method.rs
```diff
@@ -474,24 +475,21 @@ impl DidPeer {
     /// See https://identity.foundation/peer-did-method-spec/#resolving-a-did
     pub fn expand_did_peer_4(&self, did: &str) -> Result<DIDDocument, DIDPeerMethodError> {
         // Ensure long format by computing did:peer:4 short form alias
-
         let alias = Self::shorten_did_peer_4(did)?;
```
IngridPuppet, line 478: Please can you improve this comment to make it clearer that hash verification will further not be necessary?

IngridPuppet (approved):
Thank you for improving this implementation.

PR #143, opened 2024-08-06
Fix crate directory name retrieval in function `crate_name`

Hermann-Core:


generic-server/src/util/mod.rs
```diff
@@ -1,7 +1,5 @@
 #![allow(unused)]
 
 pub fn crate_name() -> String {
-    let current_dir = std::env::current_dir().unwrap();
-    let basename = current_dir.file_name().unwrap().to_str().unwrap();
-    basename.to_string()
+    std::env::var("CARGO_PKG_NAME").unwrap_or_default()
 }
```
IngridPuppet, line 5: Just a comment. I know you, do you even think this method is necessary anymore? I am okay with having it removed since the logic is no longer complex.

Hermann-Core: While we're talking, I don't really find it or the `index` module useful. We can use them in integration tests. In this way, we can also remove the `server-plugin` crate and move all the plugin logic into a `plugin` sub-module of the `generic-server`

IngridPuppet: Integration tests around the generic server will likely not be an easy affair. I suggest to keep that for later, it is definitely not the priority.

I remember we already discussed this line:
> In this way, we can also remove the server-plugin crate and move all the plugin logic into a plugin sub-module of the generic-server.
Have you tried it? You'll likely run into a dependency cycle.

Hermann-Core: We should keep that function then since it is used in the `index` module.

IngridPuppet (approved):

PR #148, opened 2024-08-11
handshake documentation folder

Blindspot22:
This  directory will contain the handshake flow which is the actual process or sequence of actions that take place during a handshake, and the handshake documentation which is the detailed description and explanation of those processes and all other documentations related to the handshacke


Blindspot22:
@francis-pouatcha please review this PR

francis-pouatcha (approved):

PR #160, opened 2024-09-06
Saving-data-in-mediator-coordination-plugin-to-avoid-repeating-operations-during-mount

ndefokou:
@IngridPuppet @Christiantyemele Pleas review this pr

### DIFF
--- a/crates/plugins/mediator-coordination/src/plugin.rs
+++ b/crates/plugins/mediator-coordination/src/plugin.rs
@@ -17,7 +17,7 @@ pub struct MediatorCoordinationPlugin {
 
 
 struct MediatorCoordinationPluginEnv {
-    _public_domain: String,
+    public_domain: String,
     storage_dirpath: String,
     mongo_uri: String,
     mongo_dbn: String,
@@ -25,7 +25,7 @@ struct MediatorCoordinationPluginEnv {
 
 /// Loads environment variables required for this plugin
 fn load_plugin_env() -> Result<MediatorCoordinationPluginEnv, PluginError> {
-    let _public_domain = std::env::var("SERVER_PUBLIC_DOMAIN").map_err(|_| {
+    let public_domain = std::env::var("SERVER_PUBLIC_DOMAIN").map_err(|_| { 
         tracing::error!("SERVER_PUBLIC_DOMAIN env variable required");
         PluginError::InitError
     })?;
@@ -46,7 +46,7 @@ fn load_plugin_env() -> Result<MediatorCoordinationPluginEnv, PluginError> {
     })?;
 
     Ok(MediatorCoordinationPluginEnv {
-        _public_domain,
+        public_domain,
         storage_dirpath,
         mongo_uri,
         mongo_dbn,
@@ -100,7 +100,7 @@ impl Plugin for MediatorCoordinationPlugin {
         };
 
         // Compile state
-        let state = AppState::from(env._public_domain.clone(), diddoc, keystore, Some(repository));
+        let state = AppState::from(env.public_domain.clone(), diddoc, keystore, Some(repository)); 
 
         // Build router
         web::routes(Arc::new(state))
--- a/crates/plugins/mediator-coordination/src/plugin.rs
+++ b/crates/plugins/mediator-coordination/src/plugin.rs
@@ -1,7 +1,7 @@
 use axum::Router;
 use keystore::filesystem::StdFileSystem;
 use plugin_api::{Plugin, PluginError};
-use mongodb::{error::Error as MongoError, options::ClientOptions, Client, Database};
+use mongodb::{ options::ClientOptions, Client, Database};
 use std::sync::Arc;
 
 use crate::{
@@ -9,9 +9,12 @@ use crate::{
     util,
     web::{self, AppState, AppStateRepository},
 };
-
 #[derive(Default)]
-pub struct MediatorCoordinationPlugin;
+pub struct MediatorCoordinationPlugin {
+    env: Option<MediatorCoordinationPluginEnv>,
+    db: Option<Database>,
+}
+
 
 struct MediatorCoordinationPluginEnv {
     _public_domain: String,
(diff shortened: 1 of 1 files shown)


crates/plugins/mediator-coordination/src/plugin.rs
```diff
@@ -55,28 +58,21 @@ impl Plugin for MediatorCoordinationPlugin {
...
         }
 
         // Check connectivity to database
+        let db = load_mongo_connector(&env.mongo_uri, &env.mongo_dbn)?;
 
-        if load_mongo_connector(&mongo_uri, &mongo_dbn).is_err() {
-            tracing::error!("could not establish connectivity with mongodb");
-            return Err(PluginError::InitError);
-        }
+        // Save the environment and MongoDB connection in the struct
```
Christiantyemele, line 73: No you can save actuall values to be reused in a different function in a self parameter

crates/plugins/mediator-coordination/src/plugin.rs
```diff
@@ -86,42 +82,45 @@ impl Plugin for MediatorCoordinationPlugin {
     }
 
     fn routes(&self) -> Router {
```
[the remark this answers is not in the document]
Christiantyemele: Check this example https://gist.github.com/rust-play/f079f88a8b79f4cbfb665f6420742780

Christiantyemele (changes requested):

crates/plugins/mediator-coordination/src/plugin.rs
ndefokou: Yes, You can. In my case,
When i store the environment (env) and database connection (db) in the self struct in mount, i can access them later in other functions by referencing self.

ndefokou: Yes, You can.In my case, 
When i store the environment (env) and database connection (db) in the self struct in mount, i can access them later in other functions by referencing self.

crates/plugins/mediator-coordination/src/plugin.rs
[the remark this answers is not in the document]
ndefokou: The self in routes refers to the same instance as in mount because i am referring to the same instance of MediatorCoordinationPlugin.

crates/plugins/mediator-coordination/src/plugin.rs
Christiantyemele: No you are missing the point try to answer this questions
- what is self
- how do i create or have a self

crates/plugins/mediator-coordination/src/plugin.rs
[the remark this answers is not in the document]
Christiantyemele: That is not possible the two functions have no relation, infact they are not aware 

> The self in routes refers to the same instance as in mount because i am referring to the same instance of MediatorCoordinationPlugin.

That is not possible the two functions have no relation, infact they are not aware of each other, i mean check my example again you will see what i mean

Christiantyemele (changes requested):
Please check the example again and compare it with yours

crates/plugins/mediator-coordination/src/plugin.rs
```diff
@@ -86,42 +82,45 @@ impl Plugin for MediatorCoordinationPlugin {
...
 
         // Load crypto identity
         let mut fs = StdFileSystem;
-        let diddoc = util::read_diddoc(&fs, &storage_dirpath).expect(msg);
-        let keystore = util::read_keystore(&mut fs, &storage_dirpath).expect(msg);
-
-        // Load connection to database
-        let db = load_mongo_connector(&mongo_uri, &mongo_dbn).expect(msg);
+        let diddoc = util::read_diddoc(&fs, &env.storage_dirpath).expect(msg);
+        let keystore = util::read_keystore(&mut fs, &env.storage_dirpath).expect(msg);
```
IngridPuppet, line 94: I think you should also do this in the mount method. What do you think?

crates/plugins/mediator-coordination/src/plugin.rs
```diff
@@ -1,17 +1,20 @@
...
 #[derive(Default)]
-pub struct MediatorCoordinationPlugin;
+pub struct MediatorCoordinationPlugin {
+    env: Option<MediatorCoordinationPluginEnv>,
+    db: Option<Database>,
+}
+
 
 struct MediatorCoordinationPluginEnv {
     _public_domain: String,
```
IngridPuppet, line 20: It is now okay to remove the leading underscore.
```suggestion
    public_domain: String,
```

IngridPuppet (changes requested):
I'll be adding to the previous discussions too.

crates/plugins/mediator-coordination/src/plugin.rs
IngridPuppet: Hi @Christiantyemele. Did you notice these are later saved in `self`? Or am I coming late?
https://github.com/adorsys/didcomm-mediator-rs/blob/de535f1277d30b12f4b94db7a1f58daf37d971ee/crates/plugins/mediator-coordination/src/plugin.rs#L73-L75

crates/plugins/mediator-coordination/src/plugin.rs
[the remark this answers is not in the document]
IngridPuppet: > Check this example https://gist.github.com/rust-play/f079f88a8b79f4cbfb665f6420742780

Hi @Christiantyemele. Do you realize in your example you are cloning the object? That's why you now have two instances. I would recommend to both of you to further check the difference between `self`, `&self`, and `&mut self`.

Here is an update of your minimal working example:

```rust
#[derive(Clone)]
#[allow(unused)]
struct Add {
    a: i32,
    b: i32
}

impl Add {
    fn add_first(&mut self, c: i32, d: i32) {
        self.a = c + d
    }
    
    fn give_from_add_first(&self) -> i32 {
        self.a
    }
}

fn main() {
    let mut add = Add {a: 4, b: 5};
    add.add_first(1, 2);
    println!("See, the logic holds with references: {}", add.give_from_add_first())
}
```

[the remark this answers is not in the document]
Christiantyemele: :sweat_smile:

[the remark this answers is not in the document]
ndefokou: Thanks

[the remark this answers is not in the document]
Christiantyemele: My bad @ndefokou i was the one wrong :sweat_smile:

Christiantyemele (approved):
Nice :+1:

crates/plugins/mediator-coordination/src/plugin.rs
ndefokou: That’s a good suggestion , but moving it to the mount method would require some refactoring of the codebase. Since the current approach isn’t causing any issues, it might be better to leave it as is for now.

IngridPuppet (approved):

PR #161, opened 2024-09-06
Iss 41 forward protocol

Christiantyemele:
https://didcomm.org/book/v2/routing


crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,121 @@
...
+struct MessageStore {
+    dirpath: String,
+}
+impl MessageStore {
+    fn persist_msg(&self, content: String, inode: String) {
+        fs::create_dir_all(&self.dirpath).unwrap();
+        let file = format!("{}/{}.json",&self.dirpath, inode,);
+        fs::write(file, content).unwrap();
+    }
+}
```
IngridPuppet, line 16: Are you persisting messages to raw files? This should be done to a MongoDB collection instead. Please inspect how things are persisted to the database in mediation coordination steps.

IngridPuppet (changes requested):

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,121 @@
...
+    // unpack encrypted payload message
+    let store = MessageStore { dirpath: store_dir_path };
+
+    let result = Message::unpack(
+        payload,
+        &state.did_resolver,
+        &state.secrets_resolver,
+        &UnpackOptions::default(),
+    )
+    .await;
```
Hermann-Core, line 34: Use an early return to immediately propagate the error if encountered. This avoids unnecessary processing within the match block, preventing wasted computation time.
```rust
let (unpack_msg, _) = Message::unpack(
    payload,
    &state.did_resolver,
    &state.secrets_resolver,
    &UnpackOptions::default(),
).await
.map_err(|_| MediationError::MessageUnpackingFailure)?;
```

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,121 @@
...
+
+                        store.persist_msg(serde_json::to_string_pretty(&re_packed_msg).unwrap(), did)
+                    }
+                }
+
+                Ok(())
+            }
+            Err(_) => Err(MediationError::MessageUnpackingFailure),
+        }
+    }
```
Hermann-Core, line 61: You should avoid directly unwrapping code that has high chances of failures. instead use an error propagation mechanism to prevent crashes and facilitate debugging.
```rust
let (re_packed_msg, _) = unpack_msg
      .pack_encrypted(
       &did,
       mediator_did,
       None,
       &state.did_resolver,
       &state.secrets_resolver,
       &PackEncryptedOptions::default(),
)
.await
.map_err(|_| MediationError::MessagePackingFailure)?;
```
Also reduce nested conditions:
```rust
if let Some(dids) = unpack_msg.to {
    for did in dids {
        let (re_packed_msg, _) = unpack_msg
            .pack_encrypted(
                &did,
                mediator_did,
                None,
                &state.did_resolver,
                &state.secrets_resolver,
                &PackEncryptedOptions::default(),
            )
            .await
            .map_err(|_| MediationError::MessagePackingFailure)?;
        store.push(re_packed_msg);
    }
}
```

Hermann-Core (changes requested):

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: yh do insignificant but still

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+                message: unpack_msg.clone(),
+                recipient_did: did,
+            };
+            message_repository
+                .store(messages)
+                .await
+                .map_err(|_| MediationError::PersisenceError)
+                .unwrap();
+            forward.push(re_packed_msg)
+        }
```
Hermann-Core, line 59: According to the [routing protocol 2.0](https://identity.foundation/didcomm-messaging/spec/v2.1/#routing-protocol-20):
> There are 3 roles in the protocol: sender, mediator, and recipient. The sender emits messages of type **forward** to the mediator. The mediator **unpacks (decrypts)** the **payload** of an encrypted forward message and passes on the result (a blob that probably contains a differently encrypted payload) to the recipient.
>> Mediator process:
>> 1. Receive ‘forward’ message.
>> 2. Retrieve service endpoint pre-configured by recipient (**next** attribute).
>> 3. Transmit payload message to service endpoint in the manner specified in the [transports] section.

The mediator will receive an encrypted payload (a forward message from the sender), decrypts it and send it to the recipient. So you need to **unpack** the message first, look at the **next** field (which is the recipient's did or key) of the **body** section of the message, and then store that did (or key) along with the **attachments** of the message.
> Normally, the payload attached to the forward message received by the mediator is transmitted directly to the receiver with no further packaging. However, optionally, the mediator can attach the opaque payload to a new forward message (appropriately anoncrypted), which then acts as a fresh outer envelope for the second half of the delivery. This rewrapping means that the “onion” of packed messages stays the same size rather than getting smaller as a result of the forward operation.
Rewrapping mode is invisible to senders, but mediators need to know about it, since they change their behavior as a result. Receivers also need to know about it, because it causes them to receive a double-packaged message instead of a singly-packaged one. The outer envelope is a forward message where to is the receiver itself.

To support this, we can modify the signature of the **`mediator_forward_process`** function to include an additional parameter that optionally enables the rewrapping process of the attached payload before storing it. To wrap the payload, you can use the `wrap_in_forward` function from the `didcomm` crate, located at `didcomm::protocols::routing::wrap_in_forward`.

Additionally, avoid directly unwrapping results whenever possible to prevent potential runtime panics and facilitate debugging.

Hermann-Core (changes requested):

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+    payload: &str,
+    state: &AppState,
+) -> Result<Vec<String>, MediationError> {
+    // unpack encrypted payload message
+    let AppStateRepository {
+        message_repository, ..
+    } = state
+        .repository
+        .as_ref()
+        .expect("Missing Persistence Layer");
```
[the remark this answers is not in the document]
Christiantyemele: This is an Option and i don't think matching on it is the best solution. i prefere a panic with a good message

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+            .pack_encrypted(
+                "did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK",
+                Some("did:web:alice-mediator.com:alice_mediator_pub"),
+                None,
+                &state.did_resolver,
+                &state.secrets_resolver,
+                &PackEncryptedOptions::default(),
+            )
+            .await
+            .unwrap();
```
IngridPuppet, line 125: Same, the original sender is not the mediator. Also, you carefully need to prepare a **forward** message. 

I recommend to clone this crate we're using: https://github.com/sicpa-dlab/didcomm-rust. Then run `routing.rs` in the `examples` folder. You'll have an idea of a complete forward flow.

You'll also find my changes to the `single_mediator()` scenario useful.

```rust
#[allow(unused_imports, dead_code)]
#[path = "../src/test_vectors/mod.rs"]
mod test_vectors;

// TODO: look for better solution
// Allows test vectors usage inside and outside crate
pub(crate) use didcomm;

use didcomm::{
    algorithms::AnonCryptAlg,
    did::resolvers::ExampleDIDResolver,
    protocols::routing::{self, try_parse_forward},
    secrets::resolvers::ExampleSecretsResolver,
    Message, PackEncryptedOptions, UnpackOptions,
};
use serde_json::json;
use test_vectors::{
    ALICE_DID, ALICE_DID_DOC, ALICE_SECRETS, BOB_DID, BOB_DID_DOC, BOB_SECRETS, MEDIATOR1_DID_DOC,
    MEDIATOR1_SECRETS,
};

#[tokio::main(flavor = "current_thread")]
async fn main() {
    single_mediator().await;
}

async fn single_mediator() {
    // --- Building message from ALICE to BOB ---
    let msg = Message::build(
        "example-1".to_owned(),
        "example/v1".to_owned(),
        json!("example-body"),
    )
    .to(BOB_DID.to_owned())
    .from(ALICE_DID.to_owned())
    .finalize();

    // --- Packing encrypted and authenticated message ---
    let did_resolver = ExampleDIDResolver::new(vec![
        ALICE_DID_DOC.clone(),
        BOB_DID_DOC.clone(),
        MEDIATOR1_DID_DOC.clone(),
    ]);

    let secrets_resolver = ExampleSecretsResolver::new(ALICE_SECRETS.clone());

    let (msg, metadata) = msg
        .pack_encrypted(
            BOB_DID,
            Some(ALICE_DID),
            None,
            &did_resolver,
            &secrets_resolver,
            &PackEncryptedOptions {
                forward: false,
                ..Default::default()
            },
        )
        .await
        .expect("Unable pack_encrypted");

    println!("Encryption metadata is\n{:?}\n", metadata);

    // -- Wrap in forward

    let msg = routing::wrap_in_forward(
        &msg,
        None,
        BOB_DID,
        &vec![MEDIATOR1_DID_DOC.clone().id],
        &AnonCryptAlg::default(),
        &did_resolver,
    )
    .await
    .expect("Unable wrapping");

    // --- Sending message by Alice ---
    println!("Alice is sending message \n{}\n", msg);

    // --- Unpacking message by Mediator1 ---
    let did_resolver = ExampleDIDResolver::new(vec![
        ALICE_DID_DOC.clone(),
        BOB_DID_DOC.clone(),
        MEDIATOR1_DID_DOC.clone(),
    ]);

    let secrets_resolver = ExampleSecretsResolver::new(MEDIATOR1_SECRETS.clone());

    let (msg, metadata) = Message::unpack(
        &msg,
        &did_resolver,
        &secrets_resolver,
        &UnpackOptions::default(),
    )
    .await
    .expect("Unable unpack");

    println!("Mediator1 received message is \n{:?}\n", msg);

    println!(
        "Mediator1 received message unpack metadata is \n{:?}\n",
        metadata
    );

    // --- Forwarding message by Mediator1 ---
    let msg = serde_json::to_string(&try_parse_forward(&msg).unwrap().forwarded_msg).unwrap();

    println!("Mediator1 is forwarding message \n{}\n", msg);

    // --- Unpacking message by Bob ---
    let did_resolver = ExampleDIDResolver::new(vec![
        ALICE_DID_DOC.clone(),
        BOB_DID_DOC.clone(),
        MEDIATOR1_DID_DOC.clone(),
    ]);

    let secrets_resolver = ExampleSecretsResolver::new(BOB_SECRETS.clone());

    let (msg, metadata) = Message::unpack(
        &msg,
        &did_resolver,
        &secrets_resolver,
        &UnpackOptions::default(),
    )
    .await
    .expect("Unable unpack");

    println!("Bob received message is \n{:?}\n", msg);
    println!("Bob received message unpack metadata is \n{:?}\n", metadata);
}
```

crates/plugins/mediator-coordination/src/model/stateful/entity.rs
```diff
@@ -22,6 +23,14 @@ pub struct Connection {
...
     pub routing_did: String,
 }
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct Messages {
+    #[serde(rename = "_id")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<ObjectId>,
+    pub message: Message,
+    pub recipient_did: String,
+}
```
IngridPuppet, line 33: Prefer a singular name.
```suggestion
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
pub struct RoutedMessage {
    #[serde(rename = "_id")]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub id: Option<ObjectId>,
    pub message: Message,
    pub recipient_did: String,
}
```

crates/plugins/mediator-coordination/src/web/mod.rs
```diff
@@ -34,21 +33,22 @@ pub struct AppState {
...
 
     // DIDComm Resolvers
-    did_resolver: LocalDIDResolver,
-    secrets_resolver: LocalSecretsResolver,
+    pub did_resolver: LocalDIDResolver,
+    pub secrets_resolver: LocalSecretsResolver,
 
     // Persistence layer
-    repository: Option<AppStateRepository>,
+    pub repository: Option<AppStateRepository>,
```
IngridPuppet, line 44: Did you really have to make these public?

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+    model::stateful::entity::Messages,
+    web::{error::MediationError, AppState, AppStateRepository},
+};
+/// mediator receives messages of type forward then it unpacks the messages and stores it for pickup
+/// the unpacked message is then repacked for further transmission.
+/// Note: Stored messages are not re_packed and must be before transmission in case of
+/// Rewrapping.
+pub async fn mediator_forward_process(
+    payload: &str,
+    state: &AppState,
```
IngridPuppet, line 13: You defined a handler, but did you register it on the unified route? How will it be accessed from the server interface? Check how other DIDComm messages are handled:
https://github.com/adorsys/didcomm-mediator-rs/blob/4d19ef5bd11c644c077c19d9dd0bec3a93a3e28a/crates/plugins/mediator-coordination/src/web/handler.rs#L15-L43

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+            let messages = Messages {
+                id: None,
+                message: unpack_msg.clone(),
+                recipient_did: did,
+            };
+            message_repository
+                .store(messages)
+                .await
+                .map_err(|_| MediationError::PersisenceError)
+                .unwrap();
```
IngridPuppet, line 57: Why create this custom error if you are not going to propagate it anyway?

IngridPuppet (changes requested):
You added a test on the happy path, i.e., when all inputs are as expected. Note that it is equally important to test some negative paths, if possible.

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: yes i agree with you, i thought of this too but there are somme points i want to bring out
- you remember the first session with @francis-pouatcha where i said the **body** field of the didcomm::Message struct does not has any **next** attribute if you check on that.

crates/plugins/mediator-coordination/src/forward/routing.rs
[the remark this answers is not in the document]
Hermann-Core: > This is an Option and i don't think matching on it is the best solution. i prefere a panic with a good message

It’s not a matter of personal preference or any other subjective reasoning, it is about following best practices for error handling. Panics should only be used in situations where the behavior is undefined or the error is truly unrecoverable. You should propagate the error to the caller, providing clear context on why the code failed. You can convert an `Option` to a `Result` with the `ok_or` or `ok_or_else` functions.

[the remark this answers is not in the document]
Christiantyemele: ok , i was'nt  aware of this conversion

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+/// mediator receives messages of type forward then it unpacks the messages and stores it for pickup
+/// the unpacked message is then repacked for further transmission.
+/// Note: Stored messages are not re_packed and must be before transmission in case of
+/// Rewrapping.
+pub async fn mediator_forward_process(
+    payload: &str,
+    state: &AppState,
+) -> Result<Vec<String>, MediationError> {
+    // unpack encrypted payload message
+    let AppStateRepository {
```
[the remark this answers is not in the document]
Christiantyemele: opps let me adjust this

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+    let (unpack_msg, _) = Message::unpack(
+        payload,
+        &state.did_resolver,
+        &state.secrets_resolver,
+        &UnpackOptions::default(),
+    )
+    .await
+    .map_err(|_| MediationError::MessageUnpackingFailure)?;
+    if unpack_msg.to.is_some() {
+        let dids = unpack_msg.clone().to.expect("to field is None");
```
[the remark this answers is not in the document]
Christiantyemele: No its not :sweat_smile:

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+                .pack_encrypted(
+                    &did,
+                    Some(mediator_did),
+                    None,
+                    &state.did_resolver,
+                    &state.secrets_resolver,
+                    &PackEncryptedOptions::default(),
+                )
+                .await
+                .expect("could not pack message: {0}");
```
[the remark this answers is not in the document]
Christiantyemele: yes i was also wondering on this scenario, then @Hermann-Core can you also give a look at this situation

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+            message_repository
+                .store(messages)
+                .await
+                .map_err(|_| MediationError::PersisenceError)
+                .unwrap();
+            forward.push(re_packed_msg)
+        }
+    }
+
+    Ok(forward)
```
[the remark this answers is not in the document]
Christiantyemele: I did this is incase of live mode being enables where forward will just give the message on pickup query

crates/plugins/mediator-coordination/src/model/stateful/entity.rs
Christiantyemele: okay i guess its more descriptive

crates/plugins/mediator-coordination/src/web/mod.rs
Christiantyemele: yes, if not i would'nt have been able to use them

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: okay :+1:

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: its fixed now

crates/plugins/mediator-coordination/src/forward/routing.rs
Hermann-Core: The content of the `body` field depends on the `type` attribute of the DIDComm message and must be a JSON object.
> body - OPTIONAL. The body attribute contains all the data and structure that are uniquely defined for the schema associated with the `type` attribute. If present, it MUST be a JSON object that conforms to [RFC 7159](https://datatracker.ietf.org/doc/html/rfc7159).

Since the `next` attribute is a String, you need to check it within the `String(String)` variant of the JSON `Value` type from `didcomm`.

Christiantyemele: Thanks for the hint :pray:

crates/plugins/mediator-coordination/src/forward/routing.rs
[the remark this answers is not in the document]
Hermann-Core: This approach is not optimal. Instead, you could use the matching `if let Some(data)` pattern. 
Additionally, this check is unnecessary because if the mediator were not the intended recipient, it would not have been able to decrypt the message. What you need to inspect here is the `next` attribute within the `body` field to get the did (or the key) of the recipient.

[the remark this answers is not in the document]
Christiantyemele: :genie_man:

crates/plugins/mediator-coordination/src/web/mod.rs
Hermann-Core: Consider using `pub(crate)` or `pub(super)` instead. This limits the visibility of the items to within the crate or module, respectively.

IngridPuppet:
> ```
> expires
> ```

You should only mind storing the messages along with their headers. The rule can be enforced at pickup time.

Hermann-Core:
> @Hermann-Core and @IngridPuppet please do you know how this can be handled
> 
> ```
> The sender can add the standard expires_time to a forward message. An additional header, delay_milli is also possible; this allows the sender to request that a mediator wait a specified number of milliseconds before delivering. Negative values mean that the mediator should randomize delay by picking a number of milliseconds between 0 and the absolute value of the number, with a uniform distribution.
> ```
> 
> as the messages are stored at the level of the mediator, i which to also know what happens to the message in the database when pickup

As Ingrid mentioned, you could store the payload along with headers. When messages are requested during the pickup, we can check these headers and perform the necessary processing before sending the attached payload. Alternatively, you can create a TODO ticket so we can handle this later as our priority now is to implement the basic functionalities first.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,132 @@
...
+    #[tokio::test]
+    /// simulate sender forwarding process
+    async fn test_mediator_forward_process() {
+        let msg: Message = Message::build(
+            "id".to_owned(),
+            "type_".to_owned(),
+            serde_json::json!("example-body"),
+        )
+        .to("did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK".to_owned())
+        .from("did:web:alice-mediator.com:alice_mediator_pub".to_owned())
```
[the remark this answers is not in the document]
Christiantyemele: Yes let me fix that

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: Yeah this really helped thanks :pray:

crates/plugins/mediator-coordination/src/web/mod.rs
Christiantyemele: :thumpsup:

Christiantyemele:
@Hermann-Core i this commit solves the issue, what do you think @IngridPuppet

crates/plugins/mediator-coordination/src/forward/ledger.rs
IngridPuppet, line 1: Please, reuse the existing infrastructure and do not introduce this file. I'll attempt to rewrite your test for the sake of illustration.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,183 @@
...
+        connection_repository,
+        ..
+    } = state
+        .repository
+        .as_ref()
+        .ok_or_else(|| MediationError::RepostitoryError)
+        .unwrap();
+
+    let body: Value = json!(payload.body.as_object());
+    let next: Vec<String> = from_value(body.get("next").unwrap().to_owned()).unwrap();
```
IngridPuppet, line 32: Beware of unwrapping things carelessly. Is the next field necessarily an array?

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,183 @@
...
+            Some(repository),
+        ));
+
+        state
+    }
+
+    #[tokio::test]
+    async fn test_mediator_forward_process() {
+        // simulate sender forwarding process
+        let did_resolver =
```
IngridPuppet, line 130: - Make you test more realistic by reusing existing APIs of the DIDComm library.
- You're using the same DID for the sender and the recipient, which is awkward.
- Create aliases (variables or functions) to make the story line more readable.
- Normally unpacking should be done by the app, not in the test. But we can fix that later.

Check this alternative:

```rust
#[cfg(test)]
mod test {
    use std::sync::Arc;

    use crate::{
        didcomm::bridge::LocalSecretsResolver,
        repository::stateful::coord::tests::{
            MockConnectionRepository, MockMessagesRepository, MockSecretsRepository,
        },
        util::{self, MockFileSystem},
        web::AppStateRepository,
    };

    use super::*;

    use did_utils::jwk::Jwk;
    use didcomm::{
        algorithms::AnonCryptAlg, protocols::routing::wrap_in_forward, secrets::SecretsResolver, Message, PackEncryptedOptions, UnpackOptions
    };
    use uuid::Uuid;

    pub fn setup() -> Arc<AppState> {
        let public_domain = String::from("http://alice-mediator.com");

        let mut mock_fs = MockFileSystem;
        let diddoc = util::read_diddoc(&mock_fs, "").unwrap();
        let keystore = util::read_keystore(&mut mock_fs, "").unwrap();

        let repository = AppStateRepository {
            connection_repository: Arc::new(MockConnectionRepository::from(vec![])),
            secret_repository: Arc::new(MockSecretsRepository::from(vec![])),
            message_repository: Arc::new(MockMessagesRepository::from(vec![])),
        };

        let state = Arc::new(AppState::from(
            public_domain,
            diddoc,
            keystore,
            Some(repository),
        ));

        state
    }

    #[tokio::test]
    async fn test_mediator_forward_process() {
        let state = &setup();

        /* Simulate sender forwarding process */

        let msg = Message::build(
            Uuid::new_v4().to_string(),
            "example/v1".to_owned(),
            json!("Hey there! Just wanted to remind you to step outside for a bit. A little fresh air can do wonders for your mood."),
        )
        .to(_recipient_did())
        .from(_sender_did())
        .finalize();

        let (msg, _) = msg
            .pack_encrypted(
                &_recipient_did(),
                Some(&_sender_did()),
                None,
                &state.did_resolver,
                &_sender_secrets_resolver(),
                &PackEncryptedOptions::default(),
            )
            .await
            .expect("Unable pack_encrypted");

        let msg = wrap_in_forward(
            &msg,
            None,
            &_recipient_did(),
            &vec![_mediator_did(state)],
            &AnonCryptAlg::default(),
            &state.did_resolver,
        )
        .await
        .expect("Unable wrap_in_forward");

        // Mediator in action
        let (payload, _) = Message::unpack(
            &msg,
            &state.did_resolver,
            &state.secrets_resolver,
            &UnpackOptions::default(),
        )
        .await
        .unwrap();

        assert!(mediator_forward_process(state, payload).await.is_ok());
    }

    pub fn _sender_did() -> String {
        "did:key:z6MkwKfDFAK49Lb9D6HchFiCXdcurRUSFrbnwDBk5qFZeHA3".to_string()
    }

    pub fn _mediator_did(state: &AppState) -> String {
        state.diddoc.id.clone()
    }

    pub fn _recipient_did() -> String {
        "did:key:z6MkfyTREjTxQ8hUwSwBPeDHf3uPL3qCjSSuNPwsyMpWUGH7".to_string()
    }

    pub fn _sender_secrets_resolver() -> impl SecretsResolver {
        let secret_id = _sender_did() + "#z6LSiZbfm5L5zR3mrqpHyL7T2b2x3afUMpmGnMrEQznAz5F3";
        let secret: Jwk = serde_json::from_str(
            r#"{
                "kty": "OKP",
                "crv": "X25519",
                "x": "ZlJzHqy2dLrDQNlV15O3zDOIXpWVQnq6VtiVZ78O0hY",
                "d": "8OK7-1IVMdcM86PZzYKsbIi3kCJ-RxI8XFKe9JEcF2Y"
            }"#,
        )
        .unwrap();

        LocalSecretsResolver::new(&secret_id, &secret)
    }
}
```

crates/plugins/mediator-coordination/src/web/handler.rs
```diff
@@ -54,7 +58,7 @@ async fn process_response_from_delegate_handler(
 
     // Pack response message
     let packed_message = match web::midlw::pack_response_message(
-        &plain_response_message,
+        &plain_response_message.unwrap(),
         &state.did_resolver,
```
IngridPuppet, line 62: Your forward delegate handler will inevitably fail because it returns a None response, which cannot be unwrapped. What you must do here is to conditionally pack the response message if there is Some, otherwise reply with an empty string.

IngridPuppet (dismissed):

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: @Hermann-Core and i discussed it and its actually not a vector but a single value

crates/plugins/mediator-coordination/src/forward/ledger.rs
Hermann-Core, line 1: What is the purpose of this file?

Christiantyemele: Opps it was for testing just forgot to remove it

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,225 @@
...
+        ..
+    } = state
+        .repository
+        .as_ref()
+        .ok_or_else(|| MediationError::RepostitoryError)
+        .unwrap();
+
+    // Check if the client's did in mediator's keylist
+
+    let next = payload.body.get("next").and_then(Value::as_str).unwrap();
```
Hermann-Core, line 31: What if the **body** entry is malformed? unwrapping is not the right way to do in this case.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    let _connection = match connection_repository
+        .find_one_by(doc! {"keylist": doc!{ "$elemMatch": { "$eq": &next}}})
+        .await
+        .unwrap()
```
Hermann-Core, line 36: Return a comprehensive error message rather than directly unwrapping.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,225 @@
...
+use mongodb::bson::doc;
+use serde_json::{json, Value};
+
+use crate::{
+
+    model::stateful::entity::RoutedMessage, web::{error::MediationError, AppState, AppStateRepository}
+};
+
+/// mediator receives messages of type forward then it unpacks the messages and stores it for pickup
+/// the unpacked message is then repacked for further transmission.
```
Hermann-Core, line 14: The description is not consistent with the actual implementation of the function.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,225 @@
...
+            );
+            return Err(response.into_response());
+        }
+    };
+
+    let attachments = payload.attachments.unwrap_or_default();
+    for att in attachments {
+        let attached = match att.data {
+            AttachmentData::Json { value: val } => val.json,
+            _ => json!(0),
```
Hermann-Core, line 52: Properly handle the case where the attachment is not a Json data.
```suggestion
            AttachmentData::Json { value: val } => val.json,
            _ => json!(Value::Null),
```

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,225 @@
...
+        };
+        message_repository
+            .store(RoutedMessage {
+                id: None,
+                message: json!(attached),
+                recipient_did: next.to_string(),
+            })
+            .await
+            .map_err(|_| MediationError::PersisenceError)
+            .unwrap();
```
Hermann-Core, line 62: Same here. why mapping the error if you are not going to propagate it?

crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
```diff
@@ -337,16 +337,18 @@ pub async fn process_plain_keylist_update_message(
...
 
-    Ok(Message::build(
-        format!("urn:uuid:{}", Uuid::new_v4()),
-        KEYLIST_UPDATE_RESPONSE_2_0.to_string(),
-        json!(KeylistUpdateResponseBody {
-            updated: confirmations
-        }),
+    Ok(
+        Message::build(
+            format!("urn:uuid:{}", Uuid::new_v4()),
```
Hermann-Core, line 342: There's a more transparent method for this.
```suggestion
            Uuid::new_v4().urn().to_string(),
```
What is the purpose of the **urn:uuid:** prefix?
cc @IngridPuppet

crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
```diff
@@ -337,16 +337,18 @@ pub async fn process_plain_keylist_update_message(
...
+            KEYLIST_UPDATE_RESPONSE_2_0.to_string(),
+            json!(KeylistUpdateResponseBody {
+                updated: confirmations
+            }),
+        )
+        .to(sender.clone())
+        .from(mediator_did.clone())
```
Hermann-Core, line 349: Why bother cloning these values here? They can be directly consumed.

crates/plugins/mediator-coordination/src/web/handler.rs
```diff
@@ -8,16 +8,30 @@ use hyper::{header::CONTENT_TYPE, StatusCode};
...
+    if message.type_ == MEDIATE_FORWARD_2_0 {
+        let response = mediator_forward_process(&state, message)
+            .await
+            .map(|_| StatusCode::ACCEPTED.into_response())
+            .map_err(|err| err);
+
+        return match response {
+            Ok(_message) => StatusCode::ACCEPTED.into_response(),
+            Err(response) => response,
+        };
```
Hermann-Core, line 32:
```suggestion
        return match mediator_forward_process(&state, message).await {
            Ok(_message) => StatusCode::ACCEPTED.into_response(),
            Err(response) => response,
        };
```

crates/plugins/mediator-coordination/src/web/mod.rs
```diff
@@ -10,15 +10,13 @@ use keystore::KeyStore;
...
-    model::stateful::entity::{Connection, RoutedMessage, Secrets},
-    util,
+    didcomm::bridge::{LocalDIDResolver, LocalSecretsResolver}, model::stateful::entity::{Connection, RoutedMessage, Secrets}, util
 };
 
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         // Unified route for all DIDComm messages
-        .route("/", post(handler::process_didcomm_message))
+        .route("/mediate", post(handler::process_didcomm_message))
```
Hermann-Core, line 19: Why did you change the route of the didcomm requests?

Hermann-Core (changes requested):

crates/plugins/mediator-coordination/src/web/mod.rs
Christiantyemele: Just to seperate and make this clearer since the "/" route already exits though for a different method

Hermann-Core: Are you aware that the mediator will interact with multiple clients? We need to ensure that all incoming requests are understood and handled correctly. Based on other implementations, mediation requests are typically directed to the "/" route.

crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
Blindspot22: If the values `sender` and `mediator_did` are being reused after being passed into the `.to()` and `.from()` methods, I think it is necessary to clone them, but if not then it is unnecessary cloning as @Hermann-Core mentioned.

crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
Christiantyemele: there is not purpose just wanting to give the id a format except you planned using

Blindspot22 (approved):
LGTM :rocket:

Blindspot22:
I also think this ticket has stayed more than expected. Maybe we should see on how to put more effort for it to move faster

Blindspot22:
Please @Christiantyemele add a brief description to this ticket

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,252 @@
...
+            let response = (StatusCode::BAD_REQUEST, RoutingError::MalformedBody.json());
+            return Err(response.into_response());
+        }
+    };
+    // Check if the client's did in mediator's keylist
+    let _connection = match connection_repository
+        .find_one_by(doc! {"keylist": doc!{ "$elemMatch": { "$eq": &next}}})
+        .await
+        .map_err(|_| MediationError::RepostitoryError)
+        .unwrap()
```
Hermann-Core, line 46: What is the necessity of mapping the error then?
```suggestion
        .map_err(|_| MediationError::RepostitoryError)?
```

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,252 @@
...
+        Some(connection) => connection,
+        None => {
+            let response = (
+                StatusCode::UNAUTHORIZED,
+                MediationError::UncoordinatedSender.json(),
+            );
+            return Err(response.into_response());
+        }
+    };
+    Ok(next.unwrap().to_string())
```
Hermann-Core, line 57: At this stage, `next` should have already been extracted.

A function should perform only one task at a time. If your function is written to retrieve the `next` field of the body, then that should be its sole purpose. Similarly, if it's meant to check a connection, it should only do that. I don't understand why you would separate these from the main function, it is not lengthy.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,252 @@
...
+async fn handler(state: &AppState, message: Message) -> Result<Message, MediationError> {
+    let AppStateRepository {
+        message_repository,
+        connection_repository,
+        ..
+    } = state
+        .repository
+        .as_ref()
+        .ok_or_else(|| MediationError::RepostitoryError)?;
+    let next = checks(&message, connection_repository).await.unwrap();
```
Hermann-Core, line 69: too much unwrapping.

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,252 @@
...
+    for attachment in attachments {
+        let attached = match attachment.data {
+            AttachmentData::Json { value: data } => data.json,
+            AttachmentData::Base64 { value: data } => json!(data.base64),
+            AttachmentData::Links { value: data } => json!(data.links),
+        };
+        message_repository
+            .store(RoutedMessage {
+                id: None,
+                message: json!(attached),
```
Hermann-Core, line 81: Be careful. The attachment should already be in Json format here.
```suggestion
                message: attached,
```

crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
Hermann-Core, line 342:
```suggestion
           Uuid::new_v4().urn().to_string(),
```

crates/plugins/mediator-coordination/src/forward/routing.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Mediator receives forwarded messages, extract the next field in the message body, and the attachments in the message
+/// then stores the attachment with the next field as key for pickup
+pub async fn mediator_forward_process(
+    state: &AppState,
+    payload: Message,
+) -> Result<Message, Response> {
+  
+    let result = handler(state, payload).await.unwrap();
+    Ok(result)
+}
```
Hermann-Core, line 27: I don't get the point of having such an empty function like this. The logic in the handler function should be placed here.

Hermann-Core (changes requested):

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: the reason why i have a handler function is to handled errors which cannot be handled in this function due to its signature

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: because i want it to panic with the mapped error, this is not something i want to catch, i don't know if you have noticed but i cannot propagate the error too (which again this is the type of error not to be handled )

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: my function is not to retrieve the next field is to make checks and act on the result

crates/plugins/mediator-coordination/src/forward/routing.rs
Hermann-Core: The client need to know that there were an error when trying to save its payload(s). You should not just panicked the code.
About the propagation you just need to map the error to a `Response` error. It’s easy to do.

crates/plugins/mediator-coordination/src/forward/routing.rs
Hermann-Core: Your function is extracting the `next` entry of the body and checking for an registered connection.

crates/plugins/mediator-coordination/src/forward/routing.rs
Hermann-Core: You just need to map the error to a `Response` type. It should be easy to do.

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: This is exactly why i want to panic the code because this error will occur only when the repository is not set which i believe should not be told to the client and if this error occours there is no need continuing as it will panics elsewhere if handled

crates/plugins/mediator-coordination/src/forward/routing.rs
Christiantyemele: yes the function is called checks

Christiantyemele: and getting next is also a check

crates/plugins/mediator-coordination/src/forward/routing.rs
Hermann-Core: Are you serious? A server must communicate all errors to the user. This why INTERNAL SERVER ERRORS exists. The client needs to be informed when something goes wrong, even if the issue is not directly related to their request.

Christiantyemele: i agree with you internal server errors exits which is what will happen in this case, are you telling me we should give user logging errors about are mediator ?, example tell the user ```repository error```, because that is exposing our system which is even and error we do not want to handle as its fatal i prefere a panic with the user having his 500 statuscode.

Blindspot22: > i agree with you internal server errors exits which is what will happen in this case, are you telling me we should give user logging errors about are mediator ?, example tell the user `repository error`, because that is exposing our system which is even and error we do not want to handle as its fatal i prefere a panic with the user having his 500 statuscode.

I think i agree with these. Why should a user be receiving logging errors about the mediator?

Hermann-Core: Why would you **"give users logging errors about our mediator"**? Logging internal details to the client is not advisable and not what I asked you to do. Instead, a generic error can be used to notify the client that their request was unsuccessful. The `map_err` is there to help you. Crashing should be reserved for unrecoverable errors, which is not the situation in this case. The server should not crash simply because a resource is inaccessible, that doesn't make sense. A server should remain responsive, even when the error is unrelated to the client, except in exceptional cases.

Christiantyemele: check again the mediator is not crashing because it repository was not set, if this error is not unrecoverable then idk what it is

Hermann-Core: The repository has already been set at this level. Your are just trying to access it. Please handle the error and go ahead.

Christiantyemele: how can you be sure it was already been set, and if that be the case then we are sure they will be no error so no matter how it is handled it will never be reached and if you insist in handling that error which i think should'nt then you contradict the point you just said the repository is already set, are'nt you seeing it my way ?

Hermann-Core: The repository is set here. At this point, the server has not been initiated, so it is safe to encounter a crash here and address the issue.
https://github.com/adorsys/didcomm-mediator-rs/blob/345b0c8fe5eef4f98fc9c96f6236d3d803cc64ba/crates/plugins/mediator-coordination/src/plugin.rs#L72-L109
In your case, you are accessing the repository to check for existing connections and store payloads. Various errors could occur, such as a temporary database issue and such. We typically respond by returning an error message to the client, advising them to try again later. If the error persists, our logs will alert us to the problem so we can resolve it. I believe this concludes the discussion on this matter. Please go ahead.

Christiantyemele: i might be wrong, i think the last changes solves the issue

Hermann-Core:
@Christiantyemele there are conflicts.

Christiantyemele:
> @Christiantyemele there are conflicts.

@Hermann-Core its fine now

Hermann-Core (approved):

ndefokou (approved):
Good work

Christiantyemele:
> Good work

Thanks

PR #163, opened 2024-09-09
feat(): ignoring stateless feature impl

Christiantyemele:


crates/plugins/mediator-coordination/src/web/coord/handler/stateful.rs
IngridPuppet, line 1: Can you auto-format this file? The trailing spaces are just noise.

Christiantyemele: auto-formatting find nothing to format,i mean no effect

Hermann-Core (changes requested):
I think you should also set **stateful** feature to the `stateful` module here. Because even if we don't enable the feature, the module will still be built. I don't think that this is the expected behavior.
https://github.com/adorsys/didcomm-mediator-rs/blob/bdc5d36a3d4ab6a75aa500b38be1872bd9e49586/crates/plugins/mediator-coordination/src/model/mod.rs#L1-L5

IngridPuppet (approved):

Christiantyemele:
> I think you should also set **stateful** feature to the `stateful` module here. Because even if we don't enable the feature, the module will still be built. I don't think that this is the expected behavior.
> 
> https://github.com/adorsys/didcomm-mediator-rs/blob/bdc5d36a3d4ab6a75aa500b38be1872bd9e49586/crates/plugins/mediator-coordination/src/model/mod.rs#L1-L5

Ok

Hermann-Core (approved):

PR #172, opened 2024-09-15
fix(entity): changed routed message type to string

Christiantyemele:


crates/plugins/mediator-coordination/src/model/stateful/entity.rs
IngridPuppet, line 1: Auto-format the file. There are unused imports.

crates/plugins/mediator-coordination/src/model/stateful/entity.rs
```diff
@@ -28,7 +28,7 @@ pub struct RoutedMessage {
     #[serde(rename = "_id")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub id: Option<ObjectId>,
-    pub message: Message,
+    pub message: Vec<String>,
     pub recipient_did: String,
```
IngridPuppet, line 32: if you want to store multiple entries, you have to convert the field name to its plural form.

IngridPuppet (changes requested):

IngridPuppet (approved):

PR #174, opened 2024-09-16
166 integrate support for peer did method in the mediator

Blindspot22:
Implementing `did:peer` method into our mediator
This approach will let us support both `did:key` and `did:peer` methods in our Mediator.

### DIFF
--- a/crates/did-utils/src/methods/peer/errors.rs
+++ b/crates/did-utils/src/methods/peer/errors.rs
@@ -62,3 +62,11 @@ impl From<DIDPeerMethodError> for DIDResolutionError {
         }
     }
 }
+
+impl std::fmt::Display for DIDPeerMethodError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self) // Customize this to format the error as desired
+    }
+}
+
+impl std::error::Error for DIDPeerMethodError {}
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -337,3 +337,19 @@ async fn test_local_did_resolver_resolves_peer_did_successfully() {
         json_canon::to_string(&expected).unwrap()
     );
 }
+
+async fn test_local_did_resolver_fails_on_malformed_peer_did() {
+    // Malformed Peer DID document
+    let malformed_peer_did_doc = r#"{
+        "id": "did:peer:malformed",
+        "authentication": [
+            {
+                "id": "did:peer:malformed#keys-1",
+                "type": "JsonWebKey2020",
+                "controller": "did:peer:malformed",
+                "publicKeyJwk": {
+                    "kty": "OKP",
+                    "crv": "Ed25519",
+                    "x": "InvalidPublicKey"
+                }
+            }
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -353,3 +353,16 @@ async fn test_local_did_resolver_fails_on_malformed_peer_did() {
                     "x": "InvalidPublicKey"
                 }
             }
+            ]
+    }"#;
+
+    let diddoc: Document = serde_json::from_str(malformed_peer_did_doc).unwrap();
+    let resolver = LocalDIDResolver::new(&diddoc);
+
+    // Resolving the malformed Peer DID
+    let did = "did:peer:malformed";
+    let resolved = resolver.resolve(did).await;
+
+    // Expect DIDNotResolved error due to malformed DID document
+    assert!(matches!(resolved.unwrap_err().kind(), ErrorKind::DIDNotResolved));
+}
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -320,3 +320,20 @@ async fn test_local_did_resolver_resolves_peer_did_successfully() {
         ],
         "service": []
     }"#;
+    // Mock Peer DID resolution by setting up a fake resolver
+    let diddoc: Document = serde_json::from_str(peer_did_doc).unwrap();
+    let resolver = LocalDIDResolver::new(&diddoc);
+
+    // Resolving Peer DID
+    let did = "did:peer:123456789abcdefghi";
+    let resolved = resolver.resolve(did).await.unwrap().unwrap();
+
+    // Expected DID Document
+    let expected: serde_json::Value = serde_json::from_str(peer_did_doc).unwrap();
+    
+    // Check if resolved DID document matches the expected document
+    assert_eq!(
+        json_canon::to_string(&resolved).unwrap(),
+        json_canon::to_string(&expected).unwrap()
+    );
+}
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -290,6 +290,7 @@ mod tests {
     }
 }
 
+#[tokio::test]
 async fn test_local_did_resolver_resolves_peer_did_successfully() {
     // Setup a sample Peer DID document
     let peer_did_doc = r#"{
@@ -338,6 +339,7 @@ async fn test_local_did_resolver_resolves_peer_did_successfully() {
     );
 }
 
+#[tokio::test]
 async fn test_local_did_resolver_fails_on_malformed_peer_did() {
     // Malformed Peer DID document
     let malformed_peer_did_doc = r#"{
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -289,3 +289,34 @@ mod tests {
         assert!(resolved.is_none());
     }
 }
+
+async fn test_local_did_resolver_resolves_peer_did_successfully() {
+    // Setup a sample Peer DID document
+    let peer_did_doc = r#"{
+        "id": "did:peer:123456789abcdefghi",
+        "authentication": [
+            {
+                "id": "did:peer:123456789abcdefghi#keys-1",
+                "type": "JsonWebKey2020",
+                "controller": "did:peer:123456789abcdefghi",
+                "publicKeyJwk": {
+                    "kty": "OKP",
+                    "crv": "Ed25519",
+                    "x": "Gfp9JkyH0SBz7c3skvRFS32NCMkR5VjkxF3aWjUhRXA"
+                }
+            }
+        ],
+        "keyAgreement": [
+            {
+                "id": "did:peer:123456789abcdefghi#keys-2",
+                "type": "JsonWebKey2020",
+                "controller": "did:peer:123456789abcdefghi",
+                "publicKeyJwk": {
+                    "kty": "OKP",
+                    "crv": "X25519",
+                    "x": "SCMU76FUIUVS7HDgIt6EBQHVuEzUjhnEqa7PA-UyH2w"
+                }
+            }
+        ],
+        "service": []
+    }"#;
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -11,6 +11,7 @@ use didcomm::{
     secrets::{Secret, SecretMaterial, SecretType, SecretsResolver},
 };
 use serde_json::json;
+use did_utils::methods::DidPeer;
 
 #[derive(Clone)]
 pub struct LocalDIDResolver {
@@ -29,6 +30,7 @@ impl LocalDIDResolver {
     }
 }
 
+
 #[async_trait]
 impl DIDResolver for LocalDIDResolver {
     async fn resolve(&self, did: &str) -> Result<Option<DIDDoc>> {
@@ -36,27 +38,40 @@ impl DIDResolver for LocalDIDResolver {
             return Ok(Some(self.diddoc.clone()));
         }
 
-        if !did.starts_with("did:key:") {
+        if did.starts_with("did:peer:") {
+            // Add logic to resolve Peer DIDs here
+            let method = DidPeer::new(); // Assuming a Peer DID resolver
+            match method.expand(did) {
+                Ok(diddoc) => Ok(Some(
+                    serde_json::from_value(json!(Document {
+                        service: Some(vec![]),
+                        ..diddoc
+                    }))
+                    .expect("Should easily convert between DID document representations."),
+                )),
+                
+            }
+        } else if did.starts_with("did:key:") {
+            let method = DidKey::new_full(true, PublicKeyFormat::Jwk);
+            match method.expand(did) {
+                Ok(diddoc) => Ok(Some(
+                    serde_json::from_value(json!(Document {
+                        service: Some(vec![]),
+                        ..diddoc
+                    }))
+                    .expect("Should easily convert between DID document representations."),
+                )),
+                Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, err)),
+            }
+        } else {
             return Err(Error::new(
                 ErrorKind::Unsupported,
                 DIDResolutionError::MethodNotSupported,
             ));
         }
-
-        let method = DidKey::new_full(true, PublicKeyFormat::Jwk);
-
-        match method.expand(did) {
-            Ok(diddoc) => Ok(Some(
-                serde_json::from_value(json!(Document {
-                    service: Some(vec![]),
-                    ..diddoc
-                }))
-                .expect("Should easily convert between DID document representations."),
-            )),
-            Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, err)),
-        }
     }
 }
+
 
 #[derive(Clone)]
 pub struct LocalSecretsResolver {
--- a/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
+++ b/crates/plugins/mediator-coordination/src/didcomm/bridge.rs
@@ -49,7 +49,8 @@ impl DIDResolver for LocalDIDResolver {
                     }))
                     .expect("Should easily convert between DID document representations."),
                 )),
-                
+                Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, Box::new(err))),
+
             }
         } else if did.starts_with("did:key:") {
             let method = DidKey::new_full(true, PublicKeyFormat::Jwk);


Blindspot22:
@Hermann-Core I think at this point I will like for us to have a small chat on this issue whenever you free.

Blindspot22:
> @Hermann-Core I think at this point I will like for us to have a small chat on this issue whenever you free tomorrow.

If you good with it?

Blindspot22:
Please @IngridPuppet
@Christiantyemele
 @ndefokou and @Hermann-Core I need review and approval on this ticket as fast as possible so i can move on to the next task which is the [Project Demo](https://github.com/adorsys/didcomm-mediator-rs/issues/165) for friday

crates/did-utils/src/methods/peer/errors.rs
```diff
@@ -62,3 +62,11 @@ impl From<DIDPeerMethodError> for DIDResolutionError {
...
     }
 }
+
+impl std::fmt::Display for DIDPeerMethodError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self) // Customize this to format the error as desired
+    }
+}
+
+impl std::error::Error for DIDPeerMethodError {}
```
Hermann-Core, line 72: You are not supposed to manually implement the `Display` Trait. It's not recommended either. Instead you can derive the `Debug` Trait which will automatically implement `Display` or better use the `thiserror` crate which will automatically implement both `Error` and `Display`.
```rust
use thiserror::Error;

#[derive(Debug, Error)]
pub enum DIDPeerMethodError {
#[error("Invalid DID")]
InvalidDid,
...
}
```

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
             return Ok(Some(self.diddoc.clone()));
         }
 
-        if !did.starts_with("did:key:") {
+        if did.starts_with("did:peer:") {
+            // Add logic to resolve Peer DIDs here
+            let method = DidPeer::new_with_format(PublicKeyFormat::Jwk); // Assuming a Peer DID resolver
+            match method.expand(did) {
+                Ok(diddoc) => Ok(Some(
+                    serde_json::from_value(json!(Document {
```
Hermann-Core, line 45: Why need to serialize the Document here?

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
...
+                        service: Some(vec![]),
```
Hermann-Core, line 46: `DidPeer` already contains a `service` field.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
...
+                        ..diddoc
+                    }))
+                    .expect("Should easily convert between DID document representations."),
```
Hermann-Core, line 49: Find a better way to handle the error without panicking.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
...
+                )),
+                Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, Box::new(err))),
+            }
+        } else if did.starts_with("did:key:") {
+            let method = DidKey::new_full(true, PublicKeyFormat::Jwk);
+            match method.expand(did) {
+                Ok(diddoc) => Ok(Some(
+                    serde_json::from_value(json!(Document {
```
Hermann-Core, line 57: Same here. Why need to serialize?

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
...
+                        service: Some(vec![]),
+                        ..diddoc
+                    }))
+                    .expect("Should easily convert between DID document representations."),
```
Hermann-Core, line 61: Please handle the error gracefully.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -272,4 +285,122 @@ mod tests {
...
+
+        // Debugging: Print both resolved and expected documents
+        println!(
+            "Resolved DID Doc: {}",
+            json_canon::to_string(&resolved).unwrap()
+        );
+        println!(
+            "Expected DID Doc: {}",
+            json_canon::to_string(&expected).unwrap()
+        );
```
Hermann-Core, line 340: Why do you print the results?

Hermann-Core (changes requested):

IngridPuppet:
> Please @IngridPuppet @Christiantyemele @ndefokou and @Hermann-Core I need review and approval on this ticket as fast as possible so i can move on to the next task which is the [Project Demo](https://github.com/adorsys/didcomm-mediator-rs/issues/165) against Friday

Hi Scott. You don't need to have this approved before working on the demo.

Christiantyemele:
@Blindspot22 i also think the review can wait for now

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
IngridPuppet: Let me jump in the conversation, cause I've been involved in these changes. There are actually two DID document types here. One from `did-utils` and the other from the `didcomm` dependency. We serialize and deserialize for easy conversion.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
IngridPuppet: Not always, and that's the problem. Because the type from the `didcomm` dependency requires the field, even if empty.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
IngridPuppet: There's no need to propagate errors from this conversion, cause it is a hack; and it is not expected to fail.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Hermann-Core: We can pause this for later. I'm pretty sure we can do without serializing and deserializing. They cause a lot of CPU and memory overheads.
> Most serialization frameworks like [serde](https://serde.rs/) define an internal data model that consists of basic types such as primitives, strings, and byte arrays. This splits the work of serializing a type into two stages: the frontend and the backend. The frontend takes some type and breaks it down into the serializable types of the data model. The backend then takes the data model types and writes them using some data format such as JSON, Bincode, TOML, etc. This allows a clean separation between the serialization of a type and the data format it is written to.
>
> Serde describes [its data model](https://serde.rs/data-model.html) in the serde book. Everything serialized with serde eventually boils down to some combination of those types!
>
> A major downside of traditional serialization is that it takes a considerable amount of time to read, parse, and reconstruct types from their serialized values.
>
> In JSON for example, strings are encoded by surrounding the contents with double quotes and escaping invalid characters inside of them:
>
> { "line": "\"All's well that ends well\"" }
          ^^                          ^ ^
> numbers are turned into characters:
>
> { "pi": 3.1415926 }
        ^^^^^^^^^
> and even field names, which could be implicit in most cases, are turned into strings:
>
> { "message_size": 334 }
  ^^^^^^^^^^^^^^^
All those characters are not only taking up space, they're also taking up time. Every time we read and parse JSON, we're picking through those characters in order to figure out what the values are and reproduce them in memory. An f32 is only four bytes of memory, but it's encoded using nine bytes and we still have to turn those nine characters into the right f32!

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Hermann-Core: Ok. We can simply `unwrap` then if it will never fail. Yes yes `expect` add more runtime code and space than unwrap. mdr

crates/did-utils/src/methods/peer/errors.rs
Blindspot22: Okay

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Blindspot22: Any Suggestions ?

Hermann-Core: Printing test results is not typically appropriate, as the primary goal is to verify that all tests pass. Printing can be useful for debugging during the development phase, but should be removed upon completion, leaving only the assertions in place.

Blindspot22: Ok

crates/did-utils/src/methods/peer/errors.rs
Blindspot22: Screenshot from 2024-09-25 09-32-26
I followed your suggestion to use `thiserror` instead of manually implementing the Display trait. However, I ran into an issue because the `crypto::errors::Error` type does not implement Display. Since `thiserror` relies on the Display trait for error formatting, I encountered the error:
```rust
error[E0599]: the method `as_display` exists for reference `&Error`, but its trait bounds were not satisfied
```
To resolve this, I would either need to manually implement the Display trait for `crypto::errors::Error` or refactor it to use `thiserror` as well, so that both `Display` and `Error` are automatically derived.

Blindspot22: @Hermann-Core What do you think?

Hermann-Core: Alternatively, you could map the `DIDPeerMethodError` to a `DIDResolutionError`, as it already implements all the necessary `Trait` via `thisError`. Then, adjust the signature of the `expand` method in `did:peer` to match the `expand` method signature in `did:key`:
```rust
pub fn expand(&self, did: &str) -> Result<DIDDocument, DIDResolutionError> {
```
This way, there would be no need to handle different errors within the `resolve` method of the `LocalDidResolver`. I believe that the expansion (which in fact is the resolution) of DIDs should always return a `DIDResolutionError` in case of errors, regardless of the method used.

Blindspot22: Ok on it

Blindspot22: Screenshot from 2024-09-25 12-41-04
 I think these solves the issue of mapping `DIDPeerMethodError` to `DIDResolutionError`

Blindspot22: @Hermann-Core Just pushed some current changes in respond to your review. Enable to check it and make sure it corresponds to what you are requesting.

Blindspot22:
I think the demo was done on Monday.
So I will appreciate to have some reviews like @Hermann-Core did or an approval.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Blindspot22: @Hermann-Core Done ✅

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,25 +37,37 @@ impl DIDResolver for LocalDIDResolver {
...
+            // Adding logic to resolve Peer DIDs
+            let method = DidPeer::new_with_format(PublicKeyFormat::Jwk); // Assuming a Peer DID resolver
+            match method.expand(did) {
+                Ok(diddoc) => Ok(Some(
+                    serde_json::from_value(json!(Document {
+                        service: Some(vec![]),
+                        ..diddoc
+                    }))
+                    .expect("Should easily convert between DID document representations."),
+                )),
```
Christiantyemele, line 50: What @Hermann-Core is saying is that you should propagate the error as such or otherwise to avoid panics

Blindspot22: Okay i get it :+1:

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Blindspot22: > Please handle the error gracefully.

Okay i get it :+1:

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
...
+            // Adding logic to resolve Peer DIDs
+            let method = DidPeer::new_with_format(PublicKeyFormat::Jwk); // Assuming a Peer DID resolver
+            match method.expand(did) {
+                Ok(diddoc) => {
+                    // Attempt to convert DID document representation, propagating errors
+                    match serde_json::from_value(json!(Document {
+                        service: Some(vec![]),
+                        ..diddoc
+                    })) {
+                        Ok(doc) => Ok(Some(doc)),
```
Christiantyemele, line 50: @Hermann-Core i don't think is an error we really want to catch

crates/did-utils/src/methods/peer/errors.rs
Hermann-Core: > I think these solves the issue of mapping `DIDPeerMethodError` to `DIDResolutionError`
Ok these have already been implemented. So only the signature of the `expand` method needs to be changed.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
             return Ok(Some(self.diddoc.clone()));
         }
 
-        if !did.starts_with("did:key:") {
-            return Err(Error::new(
+        if did.starts_with("did:peer:") {
+            // Adding logic to resolve Peer DIDs
+            let method = DidPeer::new_with_format(PublicKeyFormat::Jwk); // Assuming a Peer DID resolver
```
Hermann-Core, line 42:
```suggestion
            let method = DidPeer::with_format(PublicKeyFormat::Jwk);
```

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
...
+            match method.expand(did) {
+                Ok(diddoc) => {
+                    // Attempt to convert DID document representation, propagating errors
+                    match serde_json::from_value(json!(Document {
+                        service: Some(vec![]),
```
Hermann-Core, line 47: I recommend implementing `From<{DID method use}> for DIDDoc` directly for a more appropriate conversion of the current `Document` to `DIDDoc`. This eliminates the need to manually re-add the service field as done here, allowing for more precise field population and reducing reliance on `serde_json` for deserialization, which does not correctly parse the document resolved by Peer DID. You can refer to my implementation in the demo branch:
https://github.com/adorsys/didcomm-mediator-rs/blob/2364706ce7fd515ec1b485b34962e9af32d899fc/crates/plugins/mediator-coordination/src/didcomm/bridge.rs#L54-L59
And that's all.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
...
+                        ..diddoc
+                    })) {
+                        Ok(doc) => Ok(Some(doc)),
+                        Err(err) => Err(Error::new(
+                            ErrorKind::DIDNotResolved, 
+                            Box::new(err)
+                        )),
+                    }
+                },
+                Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, Box::new(err))),
```
Hermann-Core, line 57: Why boxing the error here? You can simply return the error.
```rust
Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, err)),
```
And since both resolutions now return the same error, you don't need to handle them separately.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
...
+                        service: Some(vec![]),
+                        ..diddoc
+                    })) {
+                        Ok(doc) => Ok(Some(doc)),
+                        Err(err) => Err(Error::new(
+                            ErrorKind::DIDNotResolved, 
+                            Box::new(err)
+                        )),
+                    }
+                },
```
Hermann-Core, line 74: See my comment about implementing `From<{DID method use}> for DIDDoc`. It will be clearer and more concise.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -36,28 +37,53 @@ impl DIDResolver for LocalDIDResolver {
...
+                Err(err) => Err(Error::new(ErrorKind::DIDNotResolved, err)),
```
Hermann-Core, line 75: The same error is returned, it can be handled at the end of the unique `match` block.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -272,4 +298,114 @@ mod tests {
...
+    // Attempt to resolve the non-existent Peer DID
+    let resolved = resolver.resolve(non_existent_peer_did).await;
+
+    // Assert that the resolver returns a DIDNotResolved error for the non-existent DID
+    assert!(matches!(
+        resolved.unwrap_err().kind(),
+        ErrorKind::DIDNotResolved
+    ));
+}
+
```
Hermann-Core, line 410: Where is the test for successful `did:peer` resolution?

Hermann-Core (changes requested):

Hermann-Core:
For the `new` constructor of `LocalDIDResolver`, I believe it isn't necessary to include a document as a parameter. @IngridPuppet, could you clarify the purpose of the argument in the `constructor here`?
https://github.com/adorsys/didcomm-mediator-rs/blob/0c57bbbff0eb972ed54869fbf1df06bc03bad69e/crates/plugins/mediator-coordination/src/didcomm/bridge.rs#L21-L28

crates/did-utils/src/methods/peer/method.rs
```diff
@@ -221,18 +221,19 @@ impl DidPeer {
...
+            return Err(DIDResolutionError::InvalidDid);
         }
 
         match did {
-            s if s.starts_with("did:peer:0") => self.expand_did_peer_0(did),
-            s if s.starts_with("did:peer:2") => self.expand_did_peer_2(did),
-            s if s.starts_with("did:peer:4") => self.expand_did_peer_4(did),
-            _ => Err(DIDPeerMethodError::UnsupportedPeerDIDAlgorithm),
-        }
+        s if s.starts_with("did:peer:0") => self.expand_did_peer_0(did).map_err(Into::into),  // Convert DIDPeerMethodError to DIDResolutionError
```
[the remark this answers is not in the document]
Blindspot22: Seriously ??
I don't think the comment affects the code in any way

crates/did-utils/src/methods/peer/method.rs
```diff
@@ -221,18 +221,19 @@ impl DidPeer {
...
+        s if s.starts_with("did:peer:2") => self.expand_did_peer_2(did).map_err(Into::into),
+        s if s.starts_with("did:peer:4") => self.expand_did_peer_4(did).map_err(Into::into),
```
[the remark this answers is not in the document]
Blindspot22: I don't think so, it's true a `fully qualified method call` is used in Rust when you need to specify which trait's method to call, usually in the case of multiple traits providing methods with the same name.

But what i have implemented rather is matching against different prefixes of the DID string (`did:peer:0`, `did:peer:2`, `did:peer:4`) and calling different methods (`expand_did_peer_0`, `expand_did_peer_2`, `expand_did_peer_4`).
Screenshot from 2024-09-26 09-57-54


So that statement doesn't apply to this scenario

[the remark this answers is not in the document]
Blindspot22: @IngridPuppet I think you should look into this statement if I'm wrong

crates/did-utils/src/methods/peer/method.rs
```diff
@@ -221,18 +221,19 @@ impl DidPeer {
...
+        _ => Err(DIDResolutionError::MethodNotSupported),  // Map UnsupportedPeerDIDAlgorithm to MethodNotSupported
```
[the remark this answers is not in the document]
Blindspot22: I don't think that's the focus of the project. We have aim on implementing and making it work. 
But i will see on removing the comments if you think it's necessary.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Blindspot22: Ok let me see to it

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Blindspot22: I think these is what you are looking for, beginning from line 302 to 352
Screenshot from 2024-09-26 10-09-50

The test itself was written with the help of @IngridPuppet

crates/did-utils/src/methods/peer/method.rs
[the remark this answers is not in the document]
Hermann-Core: I told that it is not necessary. It add no value, everyone can understand what the code does at this level. The comments are necessary to explain non obvious parts of the code.

crates/did-utils/src/methods/peer/method.rs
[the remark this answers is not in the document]
Hermann-Core: > I don't think so, it's true a `fully qualified method call` is used in Rust when you need to specify which trait's method to call, usually in the case of multiple traits providing methods with the same name.
> So that statement doesn't apply to this scenario

Read my comment again. I'm sure that you didn't understand it.

crates/did-utils/src/methods/peer/method.rs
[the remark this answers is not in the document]
Blindspot22: That's true, It's done. I just pushed a commit for that

Christiantyemele (approved):

ndefokou (approved):

Tekum-Emmanuella (approved):

PR #188, opened 2024-09-26
added default diddoc

Christiantyemele:


Hermann-Core:
Isn't this actually part of the ticket that Scott is working on?

Christiantyemele:
> Isn't this actually part of the ticket that Scott is working on?

i do not know if this is part of what he had to do, but i need to continue be it true lets say the dependency is removed

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -27,6 +27,17 @@ impl LocalDIDResolver {
             .expect("Should easily convert between DID document representations."),
         }
     }
+    pub fn default() -> Self {
+        Self {
+            diddoc: DIDDoc {
```
IngridPuppet, line 32: Prefer implementing the standard Default trait at once.

Hermann-Core:
Isn’t this ticket supposed to be deleted? What is its purpose ?

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
```diff
@@ -29,6 +29,19 @@ impl LocalDIDResolver {
     }
 }
 
+impl Default for LocalDIDResolver {
+    fn default() -> Self {
+        Self {
+            diddoc: DIDDoc {
+                id: "".to_string(),
+                key_agreement: vec![],
```
IngridPuppet, line 37: I didn't really check the related issue this morning. Now that it's done, if you would like to have the resolver work without the static DIDDoc logic, why not simply set the property behind an Option. It's better than having an empty one like in this solution.

Also, @Hermann-Core said:
> Isn’t this ticket supposed to be deleted? What is its purpose ?

It seems I'm not aware of some additional details. You didn't attend to his comment so far.

IngridPuppet (commented):
I didn't really check the related issue this morning.

crates/plugins/mediator-coordination/src/didcomm/bridge.rs
Christiantyemele: yes wrapping and Option will be another way of going through it, but is it better than implementing a default resolver? , as for @Hermann-Core comment he said the issue was part of @Blindspot22 ticket which i told him i needed in the did rotation maybe he is not up to date with the new changes

Hermann-Core: @IngridPuppet's assertion regarding the Option wrapper is a good idea too. As an alternative, `Default` is implemented for the input `Document`, you could use it:
```rust
let resolver = LocalDIDResolver::new(Document::default());
```

IngridPuppet: Sorry, I omitted to submit my previous comment last week.

Christiantyemele: > I'm actually encouraging you to do both. The default impl will therefore be easily derived.
> 
> ```rust
> #[derive(Clone, Default)]
> pub struct LocalDIDResolver {
>     diddoc: Option<DIDDoc>,
> }
> ```

But in this case might have been best but default is not implemented on DiDoc

IngridPuppet: That will indeed not be required. Option's default is None.

Christiantyemele: > @IngridPuppet's assertion regarding the Option wrapper is a good idea too. As an alternative, `Default` is implemented for the input `Document`, you could use it:
> 
> ```rust
> let resolver = LocalDIDResolver::new(Document::default());
> ```

Yeah this is also a suitable possiblility, now lets compare this have default on LocalDiDResolver i think its for descriptive in solving the issue wanna solve

IngridPuppet (approved):

PR #197, opened 2024-10-01
Fix response structures in mediator coordination protocol

Hermann-Core:


Christiantyemele (approved):

Christiantyemele:
@Hermann-Core core you have some conflicts though

ndefokou (approved):
Good

Blindspot22 (approved):
`mediate_deny` and `mediate_grant` are structured correctly, adhering to the protocol's expectations. So i think everything is good with regards to the ticket aim.

crates/plugins/mediator-coordination/src/model/stateful/coord.rs
```diff
@@ -279,22 +258,21 @@ pub struct KeylistBody {
...
     pub offset: i32,
     pub remaining: i32,
 }
+
+/// Message to grant a routing DID to a mediator.
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
-#[serde(rename_all = "snake_case")]
 pub struct MediationGrantBody {
     pub routing_did: String,
 }
```
IngridPuppet, line 278: This does not sound valid a description to me. The mediator, as the message sender, does not grant a routing DID to itself. The grant message is instead an agreement to mediate a requesting agent's messages.

IngridPuppet (changes requested):

crates/plugins/mediator-coordination/src/model/stateful/coord.rs
Hermann-Core: I don’t know how this comment appeared here, maybe a dumb copy paste. I will fix that.

IngridPuppet (approved):

PR #198, opened 2024-10-02
193 upgrade mediator coordination plugin to use database stored secrets instead of filesystem keystore

ndefokou:
@Hermann-Core @IngridPuppet Please review this PR

### DIFF
--- a/crates/plugins/mediator-coordination/src/plugin.rs
+++ b/crates/plugins/mediator-coordination/src/plugin.rs
@@ -94,8 +94,8 @@ impl Plugin for MediatorCoordinationPlugin {
         let mut fs = StdFileSystem;
         let diddoc = util::read_diddoc(&fs, &env.storage_dirpath).expect(msg);
 
-        // Load secrets from the database
-        let secret_repository = MongoSecretsRepository::from_db(&db);
+        // Load secrets directly from the database
+        let secret_repository = Arc::new(MongoSecretsRepository::from_db(&db));
         let secrets = tokio::task::block_in_place(|| {
             tokio::runtime::Handle::current().block_on(secret_repository.find_all())
         }).expect("Failed to load secrets from database");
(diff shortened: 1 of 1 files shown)


crates/plugins/mediator-coordination/src/plugin.rs
```diff
@@ -92,7 +93,15 @@ impl Plugin for MediatorCoordinationPlugin {
         // Load crypto identity
         let mut fs = StdFileSystem;
         let diddoc = util::read_diddoc(&fs, &env.storage_dirpath).expect(msg);
-        let keystore = util::read_keystore(&mut fs, &env.storage_dirpath).expect(msg);
+
+        // Load secrets directly from the database
+        let secret_repository = Arc::new(MongoSecretsRepository::from_db(&db));
+        let secrets = tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(secret_repository.find_all())
+        }).expect("Failed to load secrets from database");
```
IngridPuppet, line 101: First, you're loading secrets from the database, but you subsequently do nothing with them. Notice that the `secrets` variable remains unused. 

Second, I do not agree with the main line of your solution. I understand that this ticket is about treating the database as our new keystore solution. You should rather develop a new keystore interacting with the database upon request, not duplicate the source of truth.

Also bear in mind that secrets are added to the keystore while the server is live. You reasonably cannot rely on a logic that requires remounting the plugin to load recently added secrets.

IngridPuppet (changes requested):

crates/plugins/mediator-coordination/src/plugin.rs
ndefokou: Okay, but I am going to leave this to @Hermann-Core. It is part of what he is working on, actually, and he said he is going to do it.

IngridPuppet:
That means we can close this PR right?

Hermann-Core:
> That means we can close this PR right?

Yes

PR #236, opened 2024-11-06
231 discover feature protocol

Christiantyemele:
## This is a discussion
As from the spec [here:](https://didcomm.org/discover-features/2.0/)  we will need to have a way to know in real time what protocols our mediator supports and is willing to disclose, From our previous disscussion i think we brought out a way which could help, were we said the protocols will be loaded in a hashmap, i thought of search for the protocols to disclosed in that hashmap, I am not a 100% sure its the best way to it but for now its the only way i see, But i will need to collaborate with @Hermann-Core and @ndefokou if that be the case.
Right now i have done a version taking aside the above suggestion and used a mock, but i'ld still like the implementation to be checked

Emphasise should'nt be laid on the test for now its not complete !


Christiantyemele:
> _ No description provided. _

Hermann-Core:
> As from the spec [here:](https://didcomm.org/discover-features/2.0/) we will need to have a way to know in real time what protocols our mediator supports and is willing to disclose, From our previous disscussion i think we brought out a way which could help, were we said the protocols will be loaded in a hashmap, i thought of search for the protocols to disclosed in that hashmap, I am not a 100% sure its the best way to it but for now its the only way i see, But i will need to collaborate with @Hermann-Core and @ndefokou if that be the case. 

Protocols should be separate and not directly communicate with each other. We need a method to identify supported protocols. Currently, we have a couple of solutions:
* Reading the manifest file of the **didcomm-messaging** plugin to get the declared features, although:
  * It doesn't provide the versions of the supported protocols.
  * Some protocols may be intentionally omitted during the loading of the mediator server.
* Adding the collected protocols into the **`AppState`**, thus, the handler can access and identify the mounted protocols.

Blindspot22 (changes requested):
I will approve these after the conflicts have been solved.

Blindspot22 (approved):

crates/web-plugins/didcomm-messaging/protocols/discover-features/src/handler.rs
```diff
@@ -0,0 +1,302 @@
...
+        let message = Message::build(id, QUERY_FEATURE.to_string(), json!(body)).finalize();
+
+        match handle_query_request(message, state) {
+            Ok(_) => {
+                panic!("This should'nt occur");
+            }
+            Err(_) => {}
+        }
+    }
+}
```
Hermann-Core, line 302: Please include additional test scenarios, such as various message types from the same protocol (to verify that only the protocols are returned and only once), queries that do not match the supported protocols, bad requests, unsupported **feature-type** to see if your code behaves as expected.

Hermann-Core (changes requested):

crates/web-plugins/didcomm-messaging/protocols/discover-features/src/handler.rs
```diff
@@ -0,0 +1,302 @@
...
+                                            {
+                                                let parts: Vec<&str> =
+                                                    container.split(id).collect();
+                                                let minor = parts[1]
+                                                    .to_string()
+                                                    .chars()
+                                                    .nth(1)
+                                                    .unwrap_or_default();
+                                                let id = format!("{id}.{minor}");
+                                                disclosed_protocols.insert(id.to_string());
```
Hermann-Core, line 78: What are you trying to do here?

crates/web-plugins/didcomm-messaging/protocols/discover-features/src/handler.rs
```diff
@@ -0,0 +1,294 @@
...
+            keystore: Arc::new(MockKeyStore::new(vec![])),
+        };
+
+        let state = Arc::new(AppState::from(
+            public_domain,
+            diddoc,
+            Some(vec![
+                "https://didcomm.org/coordinate-mediation/2.0/mediate-request".to_string(),
+            ]),
+            Some(repository),
```
Hermann-Core, line 156: You can simply enter the `None` parameter if you are not using the App repository.

Hermann-Core:
Please merge the main branch to yours.

Hermann-Core (approved):

PR #219, opened 2024-10-16
adding messaging plugin container with the various messaging plugins

ndefokou:
@Hermann-Core @Christiantyemele @IngridPuppet please review this PR

PR #244, opened 2024-11-12
240 change the signature of the request handlers functions in the didcomm messaging plugin

Hermann-Core:


Christiantyemele (approved):

ndefokou (approved):
Good :+1:

crates/web-plugins/didcomm-messaging/src/web/dispatcher.rs
```diff
@@ -23,67 +21,85 @@ pub(crate) async fn process_didcomm_message(
...
-                Arc::clone(&state),
-                message,
-            )
-            .await
-        }
-        KEYLIST_QUERY_2_0 => {
-            web::handler::stateful::process_plain_keylist_query_message(Arc::clone(&state), message)
+    let response: Result<Option<Message>, Response> = match message.type_.as_str() {
+        MEDIATE_FORWARD_2_0 => {
+            forward::web::handler::mediator_forward_process(state.clone(), message)
```
IngridPuppet, line 26: It is important that handlers really have the **same** signature. There should a base error that can uniformly be turned into an axum response. That said, you should be able to have `.map_err(|e| e.into_response())` once.
```rust
let response: Result<Option<Message>, Error> = match message.type_.as_str() {
```

crates/web-plugins/didcomm-messaging/protocols/forward/src/web/handler.rs
```diff
@@ -1,14 +1,14 @@
...
 /// Mediator receives forwarded messages, extract the next field in the message body, and the attachments in the message
 /// then stores the attachment with the next field as key for pickup
 pub async fn mediator_forward_process(
-    state: &AppState,
+    state: Arc<AppState>,
     payload: Message,
-) -> Result<Message, Response> {
-    let result = handler(state, payload).await.unwrap();
+) -> Result<Option<Message>, ForwardError> {
+    let result = handler(state.clone(), payload).await.unwrap();
```
IngridPuppet, line 12: Is this `clone()` necessary?

IngridPuppet (changes requested):

Blindspot22 (approved):
LGTM :rocket:

crates/web-plugins/didcomm-messaging/protocols/forward/src/web/handler.rs
Hermann-Core: Yes. Otherwise, the `Arc` value would have been moved and we would no longer have been able to access it in the `process_response` function.

crates/web-plugins/didcomm-messaging/src/web/dispatcher.rs
Hermann-Core: The choice of decoupling the errors has been motivated by many decisions:
* The protocols are independent and designed as individual crates. Each crate becomes self-contained, and changes to one protocol’s error handling don’t affect others. Other crates wouldn't be forced to import errors that they don't need or are not aware of.
* Decoupling the errors enable us to handle errors that are specific to the corresponding protocol only. Thus, we can remove, add or modify existing protocol at any time without touching anything else.
* It would be easy to maintain since in case of error, we can immediately recognize the scope of each error type.
* With unified error type, any change would require a recompilation of each crate that uses it.

As we design a collector to gather all loaded protocols and their handlers, the need for repeatedly calling `into_response()` for each error will be eliminated. We will obtain the appropriate handler from the collector based on the message type, allowing us to call `into_response()` just once for the corresponding handler.

Hermann-Core: > I'm not asking to use the same error, but the same base error. May I know what final signature did you agree with @ndefokou for the handlers? Consider using it here.

You're referring to a general `Error` type that encompasses the errors from all protocols, right? If so, that's precisely what I'm addressing here.

IngridPuppet: Yes, exactly. Do you mean you're providing it in this PR? What is its name pls? I could not identify it. It should enforce the IntoResponse trait.

Hermann-Core: > Yes, exactly. Do you mean you're providing it in this PR? What is its name pls? I could not identify it. It should enforce the IntoResponse trait.

No I meant that’s what my previous response was trying to explain. Why I’ve chosen this error handling design instead of the shared error type.

IngridPuppet: We're still not getting each other. Let's resume the discussion on #219 when the feature branch will be updated with these changes.

IngridPuppet (approved):

PR #263, opened 2024-11-18
251 cicd review and enhancement for didcomm mediator project e3

ndefokou:
@Christiantyemele @Hermann-Core @IngridPuppet Please review this PR

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -1,0 +1,21 @@
+name: CD 
+
+on:
+  workflow_run:
+    workflows: ["Rust CI Pipeline"]  # Trigger on successful CI completion
+    types:
+      - completed
+jobs:
+  deploy:
+    runs-on: ubuntu-latest  
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: actions/setup-rust@v1
+        with:
+          rust-version: stable
+
+      # Deploy to AWS EC2 Or another instance
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -1,13 +1,12 @@
-name: CD 
+name: CD
 
 on:
-  workflow_run:
-    workflows: ["Rust CI Pipeline"]  # Trigger on successful CI completion
-    types:
-      - completed
+  push:
+    branches: [main]
+
 jobs:
   deploy:
-    runs-on: ubuntu-latest  
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,4 @@
-name: Rust CI/CD
+name: Rust CI
 
 on:
   push:
@@ -12,9 +12,9 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  # Continuous Integration Job
+
   ci:
-    name: Continuous Integration
+    name: Build and test
     runs-on: ubuntu-latest
 
     steps:
@@ -57,33 +57,3 @@ jobs:
 
     - name: Run Tests
       run: cargo nextest run --workspace --all-features
-
-  # Continuous Deployment Job
-  cd:
-    name: Continuous Deployment
-    needs: ci 
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Checkout code
-      uses: actions/checkout@v4
-
-    - name: Set up Rust
-      uses: actions-rust-lang/setup-rust-toolchain@v1
-      with:
-        toolchain: stable
-
-    - name: Build Release
-      run: cargo build --release --workspace --all-features
-
-    - name: Deploy to Production
-      env:
-        SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
-        HOST: ${{ secrets.PRODUCTION_HOST }}
-        USER: ${{ secrets.PRODUCTION_USER }}
-      run: |
-        echo "${{ secrets.SSH_PRIVATE_KEY }}" > ssh_key
-        chmod 600 ssh_key
-        scp -i ssh_key target/release/your_binary ${{ env.USER }}@${{ env.HOST }}:/path/to/your/binary
-        ssh -i ssh_key ${{ env.USER }}@${{ env.HOST }} 'systemctl restart your-service'
-        rm -f ssh_key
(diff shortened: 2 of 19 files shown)


.github/workflows/rust.yml
```diff
@@ -1,13 +1,20 @@
...
+name: Rust CI/CD
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
IngridPuppet, line 9: Create another workflow file for CD, the triggers  are generally not the same. You run CI on every open pull request, and you run CD on merge. You would not want to run CD as soon as someone creates a PR.

.github/workflows/rust.yml
```diff
@@ -35,14 +42,48 @@ jobs:
...
+        HOST: ${{ secrets.PRODUCTION_HOST }}
+        USER: ${{ secrets.PRODUCTION_USER }}
       run: |
-        cargo build --workspace --all-features
-        cargo nextest run --workspace --all-features
+        echo "${{ secrets.SSH_PRIVATE_KEY }}" > ssh_key
+        chmod 600 ssh_key
+        scp -i ssh_key target/release/your_binary ${{ env.USER }}@${{ env.HOST }}:/path/to/your/binary
+        ssh -i ssh_key ${{ env.USER }}@${{ env.HOST }} 'systemctl restart your-service'
+        rm -f ssh_key
```
IngridPuppet, line 89: Where are you deploying the server to? Aren't we supposed to have an AWS instance set up for the project? I don't get the logic here...

IngridPuppet (changes requested):

.github/workflows/rust.yml
```diff
@@ -1,13 +1,20 @@
...
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build:
-    name: Build and Test
+  # Continuous Integration Job
+  ci:
+    name: Continuous Integration
```
Hermann-Core, line 17: Why do you change the name here? Keep the previous name. CI means nothing, it is a process not a step

.github/workflows/rust.yml
```diff
@@ -35,14 +42,48 @@ jobs:
         rustflags:
 
     - name: Install Nextest test runner
-      uses: taiki-e/install-action@nextest
+      run: cargo install cargo-nextest
```
Hermann-Core, line 45: Why did you remove the previous step command ?

.github/workflows/rust.yml
```diff
@@ -35,14 +42,48 @@ jobs:
...
+
+    - name: Build Project
+      run: cargo build --workspace --all-features
+
+    - name: Run Tests
+      run: cargo nextest run --workspace --all-features
+
+  # Continuous Deployment Job
+  cd:
+    name: Continuous Deployment
```
Hermann-Core, line 63: Again this name means nothing. And as Ingrid suggested, move the CD processes to a different file.

Hermann-Core (changes requested):

.github/workflows/rust.yml
ndefokou: We are to Have

ndefokou:
@IngridPuppet @IngridPuppet You can check the CI for the CD; we are going to discuss it with Mr. Francis in our meeting.

.github/workflows/CI.yml
```diff
@@ -1,13 +1,20 @@
 name: Rust CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
IngridPuppet, line 9: Triggering the workflow on both push and pull_request events makes it run twice unnecessarily, when a PR is open, and also when it is merged. Additionally, all PRs, even when not directed to main, should be subject to these checks.

```suggestion
on: [pull_request]
```

.github/workflows/CD.yml
```diff
@@ -0,0 +1,21 @@
+name: CD 
+
+on:
+  workflow_run:
+    workflows: ["Rust CI Pipeline"]  # Trigger on successful CI completion
+    types:
+      - completed
```
IngridPuppet, line 7: A successful CI pipeline is still not sufficient to deploy anything. CD should be triggered when a PR is merged, which fires a push event to main.

```suggestion
on:
  push:
    branches: [main]
```

IngridPuppet (changes requested):

.github/workflows/CI.yml
ndefokou: Okay thanks let me change that

.github/workflows/CI.yml
```diff
@@ -1,13 +1,14 @@
 name: Rust CI
 
-on: [push, pull_request]
+on:[push, pull_request]
```
IngridPuppet, line 3: Is there a specific reason why you want to keep the push trigger here? Note that it places an unnecessary stress on pipeline resources, especially when all branches are eligible.

```suggestion
on: [pull_request]
```

IngridPuppet (changes requested):

IngridPuppet (approved):

.github/workflows/CI.yml
ndefokou: From the other PR i am seeing that the CI is done twice and i think it is because of that last CI 
It is now clear to me

IngridPuppet (approved):

.github/workflows/CD.yml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
Hermann-Core, line 13: I think there's a recent version of this action.

.github/workflows/CI.yml
```diff
@@ -35,14 +36,18 @@ jobs:
         rustflags:
 
     - name: Install Nextest test runner
-      uses: taiki-e/install-action@nextest
+      run: cargo install cargo-nextest
```
Hermann-Core, line 39: Why did you remove the previous action? The one you've replaced it with will execute at every CI pipeline run, which is not what we want.

Hermann-Core (changes requested):

.github/workflows/CI.yml
ndefokou: Okay i see .Let me fix that

.github/workflows/CD.yml
Hermann-Core: Of course there’s is a version 4. Make sure to use the latest versions of the actions you’re using.

Hermann-Core (approved):

PR #264, opened 2024-11-19
245 integration environment for didcomm mediator server deployment e6

Christiantyemele:


src/main.rs
```diff
@@ -12,7 +12,7 @@ async fn main() {
 
     // Start server
     let port = std::env::var("SERVER_LOCAL_PORT").unwrap_or("3000".to_owned());
-    let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
+    let addr: SocketAddr = format!("0.0.0.0:{port}").parse().unwrap();
```
Blindspot22, line 15: Why was the address changed here?

Blindspot22 (changes requested):

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY . .
+
+# Build the server
+RUN cargo build --release
+
+# Use a minimal image for running the server
+FROM ubuntu
+
+RUN apt update && apt install -y libpq5
```
Blindspot22, line 13: *Just a suggestion*
I think you could use this `apt update && apt install -y libpq5 && rm -rf /var/lib/apt/lists/` instead.

Why: After installing packages, it's a good practice to remove the apt cache to reduce the image size. It actually ensures that unnecessary metadata and package lists are cleaned up.

Blindspot22 (changes requested):

src/main.rs
Christiantyemele: I am removing the map from localhost to all addresses for now as localhost will not always be valid in all environments

Blindspot22: Ohhh okay

Dockerfile
Blindspot22, line 13: *Just a suggestion*
I think you could use this `apt update && apt install -y libpq5 && rm -rf /var/lib/apt/lists/` instead.

Why: After installing packages, it's a good practice to remove the apt cache to reduce the image size. It actually ensures that unnecessary metadata and package lists are cleaned up.

Blindspot22 (changes requested):

src/main.rs
stephane-segning: Could you use a env variable too for this one and default the value to 0.0.0.0?

Christiantyemele: okay

docs/environment-setup.md
```diff
@@ -0,0 +1,42 @@
...
+
+This documentation assumes you have clone the [mediator](https://github.com/adorsys/didcomm-mediator-rs) and are in the root of the project
+
+## step 1: Setup the environment variables
+Modify the ```.env``` file with the right values for the variables
+
+## Step 2: Start The Environment
+
+```sh
+ docker-compose up -d
```
Blindspot22, line 19: *Suggestion*
## Step 2: Start The Environment

```sh
 docker-compose up -d
 ```
Verify that all containers are running by executing
```sh
docker ps
```

docs/environment-setup.md
```diff
@@ -0,0 +1,42 @@
...
+```
+## Step 3: Test Connectivity
+Use a tool like Postman or curl to verify that the server is running and responding.
+```sh 
+curl -X GET http://0.0.0.0:8080/ \
+-H "Content-Type: application/json" \
+```
```
Blindspot22, line 26: *Sggestion*
## Step 3: Test Connectivity
Use a tool like Postman or curl to verify that the server is running and responding.
```sh 
curl -X GET http://0.0.0.0:8080/ \
-H "Content-Type: application/json" \
```
Expected Response: A JSON object indicating the mediator is operational or
```sh
Give maybe an expected response or something?
```

Blindspot22 (changes requested):

Blindspot22 (approved):
Wonderful :rocket:

ndefokou (approved):

.env
```diff
@@ -0,0 +1,5 @@
+SERVER_PUBLIC_DOMAIN="http://didcomm-mediator.com"
```
Hermann-Core, line 1: Have you checked if this domain name is free?

.dockerignore
```diff
@@ -0,0 +1,2 @@
+target/
+**/target/
```
Hermann-Core, line 2: The excluded files should be more than just these.

.env
```diff
@@ -0,0 +1,5 @@
...
+SERVER_LOCAL_PORT="8080"
+STORAGE_DIRPATH="crates/generic-server/target/storage"
```
Hermann-Core, line 3: Remember that the final application binary will reside in a standalone container, so using this actual storage tree is not the best solution. We could use the same directory as the application, especially since the only file produced is the mediator DID document.
```suggestion
STORAGE_DIRPATH="."
```

.env
```diff
@@ -0,0 +1,5 @@
...
+MONGO_DBN="mediator-coordination"
```
Hermann-Core, line 4: Why choose 'mediator-coordination' as the database name? A straightforward name like 'mongodb' might make more sense.

.env
```diff
@@ -0,0 +1,5 @@
...
+MONGO_URI="mongodb://localhost:27017/"
```
Hermann-Core, line 5: It's important to keep the database URL private because, in the final production, it will contain a username and password.

Hermann-Core (changes requested):

.env
Christiantyemele: untill he have a valid name we can put anything valid here for now that why we the value comes from and env file which is customisable

.env
Christiantyemele: Since this is something internal to the mediator i preferred removing it completely and external variable and giving it to the application directly

Hermann-Core:
why is this ticket closed?

PR #275, opened 2024-11-29
Add CORS layer to mediator server for cross-origin requests

Hermann-Core:


Christiantyemele (approved):
Though we have no way of testing it :smile:

IngridPuppet:
> Though we have no way of testing it 😄

Start the server and do `fetch('http://localhost:port')` in a browser console. If the request goes through then it is all fine.

src/lib.rs
```diff
@@ -2,17 +2,28 @@ pub mod plugins;
...
+};
 
 pub fn app() -> Result<(PluginContainer<'static>, Router)> {
     let mut container = PluginContainer::new();
     container.load().map_err(|e| eyre!(e))?;
 
+    let cors = CorsLayer::new()
+        .allow_origin(Any)
+        .allow_methods([Method::GET, Method::POST])
+        .allow_headers(Any);
```
IngridPuppet, line 20: Check if `Method::OPTIONS` exists. If so, please add it as well, otherwise we're going to run into issues. It goes hand in hand with `Method::POST`.

Christiantyemele:
> > Though we have no way of testing it 😄
> 
> Start the server and do `fetch('http://localhost:port')` in a browser console. If the request goes through then it is all fine.

This will be a `get` request which is allowed by default  we need to test a `post` request instead

IngridPuppet (approved):

Issue #259, opened 2024-11-13
Write tests to validate reliability features (e.g., simulated network failures, retries) E3.


chendiblessing:
Description:
Create and implement tests that validate the reliability features of the DIDComm mediator server. This includes tests for network failure simulations, retry logic, and message acknowledgment. These tests ensure that the reliability mechanisms perform as expected under controlled failure conditions.

Acceptance Criteria:

Developed tests for simulating network failures and retry logic.
Verified the functionality of message acknowledgment mechanisms.
Documented test cases and results.

Ogenbertrand:
Take a look at this please @Christiantyemele , @ndefokou, @IngridPuppet 
Thank you

Christiantyemele:
i have close the PR with a comment

PR #280, opened 2024-12-04
253 implement fault tolerance mechanisms eg retries dead letter queues circuit breakers e5

ndefokou:
@Hermann-Core @Christiantyemele @IngridPuppet


Hermann-Core:
Have you implemented the circuit breaker logic? I didn't see it.

crates/web-plugins/didcomm-messaging/protocols/forward/src/handler.rs
```diff
@@ -36,15 +43,38 @@ pub(crate) async fn mediator_forward_process(
...
+                        })
+                        .await
+                }
+            },
+            RetryOptions::new()
+                .retries(5)
+                .exponential_backoff(Duration::from_millis(100))
+                .max_delay(Duration::from_secs(1)),
+        )
+        .await;
```
IngridPuppet, line 67: Don't you see you're repeating this logic over and over? Since you're making the repository layer resilient, I believe you should create an overloaded base repository with the retry logic, making it being reused transparently. Note also that some RepositoryError's should not lead to a retry like you're assuming here. WDYT?

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
```diff
@@ -21,9 +21,11 @@ use serde_json::json;
 use shared::{
     midlw::ensure_transport_return_route_is_decorated_all,
     repository::entity::Connection,
+    retry::{retry_async, RetryOptions},
     state::{AppState, AppStateRepository},
+    CircuitBreaker::CircuitBreaker,
```
IngridPuppet, line 26: Remove unused imports.

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
```diff
@@ -85,15 +101,24 @@ pub(crate) async fn process_mediate_request(
...
-                MediationError::InternalServerError
-            })?
-            .ok_or(MediationError::InternalServerError)?;
+        let diddoc = retry_async(
+            || {
+                let did_resolver = state.did_resolver.clone();
+                let routing_did = routing_did.clone();
+
+                async move { did_resolver.resolve(&routing_did).await.map_err(|_| ()) }
+            },
```
IngridPuppet, line 110: Is there a reason why you consider our did resolver to be unreliable?

crates/web-plugins/didcomm-messaging/protocols/pickup/src/handler.rs
```diff
@@ -489,7 +557,7 @@ mod tests {
             .await
             .unwrap_err();
 
-        assert_eq!(error, PickupError::MissingClientConnection);
+        assert_eq!(error.to_string(), "Failed to retrieve client connection");
```
IngridPuppet, line 560: Just add a new assertion without removing the pre-existing one that was actually testing for the type of the error.

crates/web-plugins/didcomm-messaging/shared/src/lib.rs
```diff
@@ -1,5 +1,7 @@
+pub mod CircuitBreaker;
```
IngridPuppet, line 1: Traditionally, modules follow the snake case.

ndefokou:
@IngridPuppet I wasn't on the right tack i am not supposed to touch the handler files i was to add those logic in the plugin file 
Screenshot from 2024-12-17 08-37-11

ndefokou:
So i am going to remove all what is in the handler and implement in the plugin

crates/web-plugins/didcomm-messaging/protocols/forward/src/handler.rs
```diff
@@ -24,6 +26,11 @@ pub(crate) async fn mediator_forward_process(
         .as_ref()
         .ok_or_else(|| ForwardError::InternalServerError)?;
 
+    let circuit_breaker = state.circuit_breaker.clone();
+    if circuit_breaker.is_open() {
+        return Err(ForwardError::CircuitOpen);
+    }
```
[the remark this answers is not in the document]
ndefokou: Okay

crates/web-plugins/didcomm-messaging/shared/src/retry.rs
```diff
@@ -0,0 +1,75 @@
+use std::time::Duration;
```
Blindspot22, line 1: Wow you already implemented the retry_logic here :+1: 
That makes less work for me on this [ticket](https://github.com/adorsys/didcomm-mediator-rs/issues/259)

Blindspot22 (approved):
I think it's good to go :rocket:

crates/web-plugins/didcomm-messaging/shared/src/retry.rs
ndefokou: :smile: great

crates/web-plugins/didcomm-messaging/protocols/forward/src/handler.rs
```diff
@@ -15,60 +17,108 @@ pub(crate) async fn mediator_forward_process(
...
         .ok_or(ForwardError::InternalServerError)?;
 
-    let next = match checks(&message, connection_repository).await.ok() {
-        Some(next) => Ok(next),
-        None => Err(ForwardError::InternalServerError),
-    };
+    let next = checks(
+        &message,
+        &repository.connection_repository,
+        state.circuit_breaker.get(MEDIATE_FORWARD_2_0).as_deref(),
```
Blindspot22, line 35: Here, if the circuit breaker is not correctly configured, the fallback `(None)` could skip validation. This could unintentionally bypass safety checks.

`suggestion`
Add logging or a fallback mechanism that fails securely if the circuit breaker is misconfigured.

Blindspot22 (changes requested):

crates/web-plugins/didcomm-messaging/protocols/forward/src/handler.rs
Hermann-Core: What do you mean? The circuit has nothing to do with the checks logic. Take a look at the `checks` function implementation.

Blindspot22: I was just suggesting that we enhanc the fallback logic to include:

    * Secure failure when the circuit breaker is unavailable or misconfigured.
    
    * Logging to indicate that a fallback path was used, which could help debug potential issues in production.
    
But it's fine, we can implement those in further time to come, It's just an amelioration.

Blindspot22 (approved):

IngridPuppet (approved):

PR #283, opened 2024-12-05
Intergration env support

Christiantyemele:
This is the architecture followed for the mediator delployment: current implementation can be found on [page2](https://drive.google.com/file/d/1EUW6UqAAo5YsR073vdLrj1DWVnYix5MV/view?usp=sharing) here 
- [mediator](https://didcomm-mediator.eudi-adorsys.com/)


Christiantyemele:
Suggested edit:
```diff
diff --git a/.env b/.env
index e2af3f8..2952ea7 100644
--- a/.env
+++ b/.env
@@ -2,4 +2,4 @@ SERVER_PUBLIC_DOMAIN="http://didcomm-mediator.com"
 SERVER_LOCAL_PORT="8080"
 STORAGE_DIRPATH="crates/generic-server/target/storage"
 MONGO_DBN="mediator-coordination"
-MONGO_URI="mongodb://0.0.0.0:27017"
+MONGO_URI="mongodb://localhost:27017"

```

Blindspot22 (approved):
Great :+1: good to go.
I think, you should squash some of the commits into a single commit or segment, because some of them don't make sense individually and a couple of repetitions.
e.g 12 commits bearing the same name :)

ndefokou:
Fixed the conflicts and after that i will approve the ticket

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Pre-cache dependencies and build the project
+RUN cargo fetch && cargo build --release
+
+# Dynamically mount source code during build for easy updates
+VOLUME /app
+
+# Ensure the binary is built and available
+CMD cargo build --release && \
+    cp target/release/didcomm-mediator /usr/local/bin/didcomm-mediator
```
Hermann-Core, line 19: Please leverage [build caching](https://docs.docker.com/build/cache/optimize/) to speed up the build process. I also suggest using the alpine version of the builder image as it will enable us to easily construct the final application image with an alpine base, resulting in the lowest possible image size.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY --from=builder /app/target/release/didcomm-mediator /usr/local/bin/didcomm-mediator
+
+# Copy runtime configurations
+COPY .env .env
+
+# Expose the port the app will listen on
+EXPOSE 3000
+
+# Set the entrypoint for the application
+ENTRYPOINT ["didcomm-mediator"]
```
Hermann-Core, line 42: As mentioned in the previous comment, we could use the Alpine image as our application base image. Additionally, it is advisable to create a new non-privileged user for the application to run under for security purposes.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: "3.8"
```
Hermann-Core, line 1: `version` is now deprecated. remove it to suppress the warning.
https://forums.docker.com/t/docker-compose-yml-version-is-obsolete/141313

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+  mediator:
+    build: 
+      context: .
+    container_name: didcomm-mediator
+    ports:
+      - "8080:8080"
+    env_file:
+      - .env
+    depends_on:
+      - mongodb
```
Hermann-Core, line 12: better
```suggestion
    depends_on:
      mongodb:
        condition: service_healthy
```

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+  mongodb:
+    image: mongo:latest
+    container_name: mongodb
+    env_file:
+      - .env
+    volumes:
+      - mongo-data:/data/db
+    networks:
+      - mediator-network
+
```
Hermann-Core, line 25: Add a health check condition to ensure the service is ready to work.
e.g.
```yaml
    healthcheck:
      test: ["CMD", "mongo", "--version"]
      interval: 10s
      timeout: 5s
      retries: 5
```

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      - mongo-data:/data/db
+    networks:
+      - mediator-network
+   
+networks:
+  mediator-network:
+    driver: bridge
+
+volumes:
+  mongo-data:
```
Hermann-Core, line 31: I do not see where you define secrets. Does this imply that the database will be accessed without any protection?

Hermann-Core (changes requested):

.gitignore
```diff
@@ -27,7 +27,7 @@ Cargo.lock
 
 # Environment variables files 
 .env.example
-.env
```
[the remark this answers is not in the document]
IngridPuppet: I'm tired of PRs changing this logic 🤦 Also note that `.env.example` SHOULD be visible, i.e., removed from `.gitignore`.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+target/
+**/target/
+.idea/
```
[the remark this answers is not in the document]
Christiantyemele: okay

.gitignore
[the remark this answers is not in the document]
Christiantyemele: okay

docker-compose.yml
Christiantyemele: ok

docker-compose.yml
Christiantyemele: true its an issue

.gitignore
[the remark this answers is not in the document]
Hermann-Core: > I'm tired of PRs changing this logic 🤦 Also note that `.env.example` SHOULD be visible, i.e., removed from `.gitignore`.

Could you tell why the `.env.example` should be visible ? I think we don’t need it anymore

chendiblessing (approved):

stephane-segning (changes requested):
It's missing, the storage path in the deployment. There should be a PVC there.

Christiantyemele:
> It's missing, the storage path in the deployment. There should be a PVC there.

where do you mean?, cause there is a pvc in the database deployment which is the only service requiring a storage medium ?

mediator-charts/templates/mongodb-deployment.yaml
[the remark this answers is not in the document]
Christiantyemele: > Please use https://artifacthub.io/packages/helm/bitnami/mongodb instead of rewriting the mongo deployment files.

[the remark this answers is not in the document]
Christiantyemele: > Please use https://artifacthub.io/packages/helm/bitnami/mongodb instead of rewriting the mongo deployment files.

i don't see the issue with the mongo deployment files since they are already written and works fine

mediator-charts/templates/secrets.yaml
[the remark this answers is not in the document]
Christiantyemele: i though this file will stored in a special way ?, cause even reading from and env file, the secrets are still in text form there

mediator-charts/templates/mongodb-deployment.yaml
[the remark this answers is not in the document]
stephane-segning: The problem is the fact, that you might make mistakes. Bitnami are known for the quality of their work. Please don't reinvent the wheel.

mediator-charts/templates/secrets.yaml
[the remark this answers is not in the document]
stephane-segning: Reading from the env is still in clear text, sure. But env are in the machine already. So we secure the access to the machine and ensure we cannot have those secrets through any other way.

stephane-segning:
> > It's missing, the storage path in the deployment. There should be a PVC there.
> 
> where do you mean?, cause there is a pvc in the database deployment which is the only service requiring a storage medium ?

Yeah, so the mediator write in `.txt` the content of the DID. Right? So, what happens if a pod restarts?

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /app
+
+# Set the storage directory path
+ENV STORAGE_DIRPATH="./storage"
+
+# Copy the built binary from the builder stage
+COPY --from=builder /app/target/release/didcomm-mediator /usr/local/bin/didcomm-mediator
+
+# Copy runtime configurations
+COPY .env .env
```
stephane-segning, line 36: Do you want to have a .env into your image? Why not leverage env variables?

stephane-segning (changes requested):

Christiantyemele:
> It's missing, the storage path in the deployment. There should be a PVC there.

exact

Dockerfile
Hermann-Core: No. it is not supposed to bring troubles. It is supposed to leverage caches to speed up the build process.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Dynamically mount source code during build for easy updates
+VOLUME /app
+
+# Ensure the binary is built and available
+CMD cargo build --release && \
+    cp target/release/didcomm-mediator /usr/local/bin/didcomm-mediator
+
+# Stage 2: Runtime
+FROM ubuntu:22.04
```
Hermann-Core, line 22: An image of 200MB just to run our application is too large. We can get really light images by statically linked our executable (e.g. using the **musl libc** or **crt-static** `rustc` feature) and base images such as `FROM scratch`,  `Distroless` or `alpine`.

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    build: 
+      context: .
+    container_name: didcomm-mediator
+    ports:
+      - "8080:8080"
+    env_file:
+      - .env  # Reference the .env file
+    # depends_on:
+      # mongodb:
+      #   condition: service_healthy
```
Hermann-Core, line 12: Why have you commented out the condition?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      - .env  # Reference the same .env file
+    volumes:
+      - mongo-data:/data/db
+    networks:
+      - mediator-network
+    # healthcheck:  
+    #   test: ["CMD", "mongo", "--version"]
+    #   interval: 10s  
+    #   timeout: 5s  
+    #   retries: 5
```
Hermann-Core, line 29: Why have you commented out the service healthcheck?

Hermann-Core (changes requested):

Dockerfile
Christiantyemele: I did this when i how the deployment would look like but i guest its not usefull again

Dockerfile
Christiantyemele: i refactor the final image to 33.4MB :+1:

docker-compose.yml
Christiantyemele: fixed

docker-compose.yml
Christiantyemele: fixed

mediator-charts/templates/mongodb-deployment.yaml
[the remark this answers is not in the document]
Christiantyemele: okay

mediator-charts/templates/secrets.yaml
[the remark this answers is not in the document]
Christiantyemele: the secrets are now kept on aws secret manager

stephane-segning (commented):
Good work. At last, I have some questions about the Dockerfile. How come we have into it the x86 only? It means if I built on top of a Mac Processor, I won't be capable of building ?

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Dummy build to cache dependencies
+RUN mkdir src && echo "fn main() {}" > src/main.rs && \
+    cargo build --release --target x86_64-unknown-linux-musl
+
+# Copy actual source code
+COPY src ./src
+
+# Build the project and strip the binary
+RUN cargo build --release --target x86_64-unknown-linux-musl && \
```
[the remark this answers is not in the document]
Christiantyemele: yh this is not something we want

Christiantyemele:
@Hermann-Core as for using mount-binds and mount-cache to increase build time of the image i found some difficulties doing that but i think it should not be a reason for this ticket to stay open any longer as it only increase image build time, we can create a refactoring issue for that in future, this is suggestion on this.

stephane-segning (approved):

ndefokou (approved):
Good :+1:

Hermann-Core (approved):

Issue #254, opened 2024-11-13
Set up message delivery guarantees and failure handling E5.


chendiblessing:
***Description:***
Establish robust message delivery guarantees to ensure reliable communication. Implement mechanisms for message acknowledgment and retries to achieve either at-least-once or exactly-once delivery semantics. Define strategies for handling message failures and ensure the server can recover and resend messages as needed.

***Acceptance Criteria:***

- Implemented acknowledgment mechanisms for message delivery.
- Configured retry logic and failure handling strategies.
- Documented delivery guarantees (e.g., at-least-once, exactly-once).

PR #284, opened 2024-12-06
259 write tests to validate reliability features eg simulated network failures retries e3

Blindspot22:
1. In `CI.yml`:

    . Added a separate `reliability-tests` job to isolate tests for retry logic and acknowledgment.
    . Used the `--filter reliability` flag in the `cargo nextest run` command to run only the reliability-related tests.
    . Cached the `target/cargo-nextest` directory to speed up future runs.

2. In `CD.yml`:

    . Added a trigger (`on.workflow_run`) to ensure deployment only happens after the `CI` workflow completes successfully.
    . Added an `if` condition to proceed with deployment only if the `CI` workflow's conclusion is `success`.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -1,8 +1,10 @@
 name: CD
 
 on:
-  push:
-    branches: [main]
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
 
 jobs:
   deploy:
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
+    - name: Checkout code
+      uses: actions/checkout@v3
 
       - name: Set up Rust
         uses: actions/setup-rust@v1
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -16,9 +16,9 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v3
 
-      - name: Set up Rust
-        uses: actions/setup-rust@v1
-        with:
-          rust-version: stable
+    - name: Set up Rust
+      uses: actions/setup-rust@v1
+      with:
+        rust-version: stable
 
       # Deploy to AWS EC2 Or another instance
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -8,6 +8,8 @@ on:
 
 jobs:
   deploy:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    name: Deploy
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -21,4 +21,7 @@ jobs:
       with:
         rust-version: stable
 
-      # Deploy to AWS EC2 Or another instance
+    - name: Deploy to AWS EC2
+      run: |
+        echo "Deploying application to AWS EC2..."
+        # Add your deployment script here
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,3 +51,10 @@ jobs:
 
     - name: Run Tests
       run: cargo nextest run --workspace --all-features
+
+  reliability-tests:
+    name: Reliability Tests
+    runs-on: ubuntu-latest
+    needs: ci
+
+   
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -66,4 +66,8 @@ jobs:
       with:
         toolchain: stable
 
-    
+    - name: Install Nextest test runner
+      uses: taiki-e/install-action@nextest
+
+    - name: Run Reliability Tests
+      run: cargo nextest run --workspace --all-features --filter reliability
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,4 +57,13 @@ jobs:
     runs-on: ubuntu-latest
     needs: ci
 
-   
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Rust
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: stable
+
+    
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,6 +24,7 @@ jobs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
+          target/cargo-nextest/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
         restore-keys: |
           ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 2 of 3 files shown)


test.rs
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            Ok("Simulated success".to_string())
+        }
+    };
+
+    let result = send_message_with_retries("test-message", mock_network).await;
+
+    assert!(result.is_ok(), "Expected success after retries");
+    assert_eq!(attempts, 3, "Expected 3 attempts before success");
+}
```
Christiantyemele, line 119: i understand what this code is all about since it is not actually testing what we have implemented as ought to do

.github/workflows/CD.yml
```diff
@@ -1,20 +1,27 @@
...
+    - name: Set up Rust
+      uses: actions/setup-rust@v1
+      with:
+        rust-version: stable
 
-      # Deploy to AWS EC2 Or another instance
+    - name: Deploy to AWS EC2
+      run: |
+        echo "Deploying application to AWS EC2..."
+        # Add your deployment script here
```
Christiantyemele, line 27: i really do not think this issue has something to do with the CI

Christiantyemele (dismissed):

Blindspot22:
> Hello @Blindspot22 - Regarding your changes on the pipeline sequence, especially on how CD is triggered, please could you sync with @ndefokou ? From memory, he was the one who wrote it in the first place and I anticipate you'll have an interesting exchange on why these changes are problematic.

Okay thanks @IngridPuppet

test.rs
Blindspot22: I have modified my tests, Please check them again

.github/workflows/CD.yml
Blindspot22: I was kind of confused before on how i was supposed to make sure this tests are emphasize, So i thought the place was to make sure it is included in the CD. Anyways, I have made alote of new changes. Please review again

test.rs
```diff
@@ -0,0 +1,228 @@
...
+    let statuses = mediator.delivery_status.lock().await;
+    let processed_ids: Vec<_> = statuses.iter().map(|(id, _)| id.clone()).collect();
+
+    assert_eq!(
+        processed_ids, message_ids,
+        "Expected messages to be processed in order: {:?}, but got {:?}",
+        message_ids, processed_ids
+    );
+}
+
```
Christiantyemele, line 228: This test file is in no crate, how do i  run it ?

Christiantyemele (dismissed):

test.rs
Blindspot22: A simple `cargo test` will run all the tests including these

src/tests/test.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    // Check that messages are processed in the correct order
+    let statuses = mediator.delivery_status.lock().await;
+    let processed_ids: Vec<_> = statuses.iter().map(|(id, _)| id.clone()).collect();
+
+    assert_eq!(
+        processed_ids, message_ids,
+        "Expected messages to be processed in order: {:?}, but got {:?}",
+        message_ids, processed_ids
+    );
```
Christiantyemele, line 225: this won't still work, either you have a test folder at the root or just a test file in the src directory for example,
i propose you just write and intergration test with test dir at the root, or you could still do otherwise with a simple unit test it both works

Christiantyemele (changes requested):

src/tests/test.rs
```diff
@@ -0,0 +1,226 @@
+use tokio::time::{sleep, Duration};
+use tokio::sync::{futures, Mutex};
+use uuid::Uuid;
```
IngridPuppet, line 3: Regarding the imports to your test module, I do not seem to understand what you will be testing. Can you help me answer that? I'm not being rhetoric, it's a genuine question.

IngridPuppet (commented):
I have

src/tests/test.rs
Blindspot22: Kk

src/tests/test.rs
IngridPuppet: Great that you're testing a lot of things. Please enlighten me on one of them. Let us consider the test with the description "Validate retry mechanism under network interruptions." I have two questions:
- Which entity in the test is subject to network interruptions that would prevent access to it?
- Which entity should implement a retry mechanism under these conditions?

Blindspot22: The entity subject to network interruptions is the DIDComm mediator server itself or any intermediary network-dependent service it relies on (e.g a database or external service). In the context of the test, this would simulate scenarios where:

    . The client agent (the test environment) attempts to send a message to the mediator server but experiences timeouts or dropped connections.
    . The mediator server itself might lose connectivity to an upstream service (e.g, a cloud-based DID registry).

Blindspot22: The entity that initiates the communication in this case, the `client agent` should implement the retry mechanism. This is because:

    The client is responsible for ensuring the message is delivered successfully to the mediator server.
    If the server is temporarily unavailable, the client should attempt to resend the message after a delay or a backoff period.
For example, If the client sends a message and receives no acknowledgment from the mediator, the client should retry according to a predefined policy (e.g, exponential backoff).

IngridPuppet: You are listing two options:
- The client agent (the test environment) attempts to send a message to the mediator server but experiences timeouts or dropped connections.
- The mediator server itself might lose connectivity to an upstream service (e.g, a cloud-based DID registry).

This ticket is about writing tests to validate reliability features, which basically implies those implemented by the mediator to remain reliable. In the first case, like you said yourself, it is the client implementing reliability features for the situation, not the mediator. And you're not expected to test features you're implementing yourself to simulate the actions of a client.

With that I believe you can discard Option 1 and focus on Option 2, so the actual reliability features assumed implemented by the mediator are tested. What do you think?

Blindspot22: Sure. Thanks for this enlightenment :+1: 
I think I'm good to go now

Blindspot22: Hello @IngridPuppet please check this again, to ensure it corresponds with your request

src/tests/reliability_tests.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    #[tokio::test]
+    async fn test_mediator_retries_on_upstream_failure() {
+        // Setup mock upstream service
+        let _mock = setup_mock_service();
+
+        // Mediator client
+        let client = Client::new();
+        let url = format!("{}/upstream", server_url());
+
```
Christiantyemele, line 16: this is not an endpoint of the mediator, so this will not make any request to it

src/tests/reliability_tests.rs
```diff
@@ -0,0 +1,41 @@
+#[cfg(test)]
+mod tests {
+    use crate::mediator::fetch_with_retries;
+    use crate::tests::mock_service::setup_mock_service;
```
Christiantyemele, line 4: place the `tests` directory at the root out of `src` if you want to write integration test.

Christiantyemele:
you can assign me too this if you need help

Blindspot22:
> you can assign me too this if you need help

Yeahh i think a bit of help will do. This ticket is taking more time than required. :)

src/tests/reliability_tests.rs
Blindspot22: Yep

src/tests/reliability_tests.rs
Blindspot22: Ohh yeah, Sorry i was to change that after doing all my proper modifications

Christiantyemele:
this issue is a sub ticket from https://github.com/adorsys/didcomm-mediator-rs/issues/254 and should be closed

PR #287, opened 2024-12-18
feat(): validated oob invitation

Christiantyemele:


ndefokou (approved):

crates/web-plugins/oob-messages/src/models.rs
```diff
@@ -74,20 +79,21 @@ impl OobMessage {
...
     }
 
-    fn serialize_oob_message(oob_message: &OobMessage, url: &str) -> Result<String, String> {
+    fn serialize_oob_message(oob_message: &OobMessage, url: &str) -> Result<Ooburl, String> {
         let plaintext =
             to_string(oob_message).map_err(|e| format!("Serialization error: {}", e))?;
         let encoded_jwm = Base64Url.encode(plaintext.as_bytes());
 
-        Ok(format!("{}?_oob={}", url, encoded_jwm))
+        let ooburl = format!("{}/_oob{}", url, encoded_jwm);
```
Hermann-Core, line 87: You changed the format of the oob_url. Is there any reason to that?

crates/web-plugins/oob-messages/src/models.rs
```diff
@@ -112,7 +118,7 @@ where
 
     let did = diddoc.id.clone();
     let oob_message = OobMessage::new(&did);
-    let url: &String = &format!("{}:{}", server_public_domain, server_local_port);
+    let url: &String = &format!("http://{}", server_public_domain);
```
Hermann-Core, line 121: appending the http scheme is not necessary here. The `server_public_domain` environment variable will likely embed the right scheme.

crates/web-plugins/oob-messages/src/plugin.rs
```diff
@@ -53,17 +48,13 @@ impl Plugin for OOBMessages {
...
-            &env.server_local_port,
-            &env.storage_dirpath,
-        )
-        .map_err(|err| {
-            PluginError::InitError(format!(
-                "Error retrieving or generating OOB invitation: {err}"
-            ))
-        })?;
+        let oob_inv =
+            retrieve_or_generate_oob_inv(&mut fs, &env.server_public_domain, &env.storage_dirpath)
```
Hermann-Core, line 52: You removed the server_local_port ?

crates/web-plugins/oob-messages/src/web.rs
```diff
@@ -4,18 +4,24 @@ use crate::{
...
 use std::sync::Arc;
 
 pub(crate) fn routes(state: Arc<OOBMessagesState>) -> Router {
+    let invitation_path = format!("/_{}", &state.oobmessage);
     Router::new() //
         .route("/oob_url", get(handler_oob_inv))
         .route("/oob_qr", get(handler_oob_qr))
         .route("/", get(handler_landing_page_oob))
+        // handle oob invitation to invitation message
+        .route(&invitation_path, get(decode_oob_inv))
```
Hermann-Core, line 17: I dont understand what the new route is and why you added it

Hermann-Core (changes requested):

crates/web-plugins/oob-messages/src/models.rs
Christiantyemele: yes i want to use the oob jwm as a routing path to the out band message and the previous format contains query parameters which will be misinterpreted if used as route path

crates/web-plugins/oob-messages/src/models.rs
Christiantyemele: yes just realized too

crates/web-plugins/oob-messages/src/plugin.rs
Christiantyemele: yes all this should be part of the SERVER_PUBLIC_DOMAIN as in prod we won't this anymore

crates/web-plugins/oob-messages/src/web.rs
Christiantyemele: The qr_code and and oob url have no handled route, what i mean is if you can the oob qr_code you get nothing even the oob_url takes you no where as the route has not been handled and as such you can have its content without decoding it, that's where this route comes in

crates/web-plugins/oob-messages/src/models.rs
```diff
@@ -39,13 +39,14 @@ use std::io::Result as IoResult;
...
 #[derive(Debug, Serialize, Deserialize, Clone)]
-struct OobMessage {
+pub(crate) struct OobMessage {
     #[serde(rename = "type")]
     oob_type: String,
     id: String,
     from: String,
     body: Body,
 }
+type Ooburl = String;
```
IngridPuppet, line 49:
```suggestion

type OobUrl = String;
```

crates/web-plugins/oob-messages/src/models.rs
```diff
@@ -249,15 +266,23 @@ mod tests {
...
-        assert!(oob_url.contains("_oob="));
+        assert!(oob_url.starts_with(&format!("{}/_oob", url)));
+        assert!(oob_url.contains("_oob"));
+    }
+    #[test]
+    fn test_deserialize_oob_message() {
+        let did = "did:key:z6MkfyTREjTxQ8hUwSwBPeDHf3uPL3qCjSSuNPwsyMpWUGH7#z6LSbuUXWSgPfpiDBjUK6E7yiCKMN2eKJsXn5b55ZgqGz6Mr";
+        let serialized_msg = test_serialize_oobmessage();
+        let oobmessage: Vec<&str> = serialized_msg.split("/_oob").collect();
+        let oobmessage = oobmessage.get(1).unwrap().to_string();
```
IngridPuppet, line 277: Note that `_oob` must be a query parameter in a compliant OOB URL.

crates/web-plugins/oob-messages/src/plugin.rs
```diff
@@ -74,8 +65,12 @@ impl Plugin for OOBMessages {
         })?;
 
         self.env = Some(env);
+        let oobmessage: Vec<&str> = oob_inv.split("/_").collect();
+        let oobmessage = oobmessage.get(1).unwrap_or(&"").to_string();
+
         self.state = Some(OOBMessagesState {
             filesystem: Arc::new(Mutex::new(fs)),
+            oobmessage,
         });
```
IngridPuppet, line 74: - I don't see the point of extracting something like "oob=abcd...xyz", when you already have the full string...
- Did you run `cargo clippy` on your changes? I feel like you can use `unwrap_or_default` here.

crates/web-plugins/oob-messages/src/web/handler.rs
```diff
@@ -1,41 +1,66 @@
...
+        <html lang="en">
+            <style>
+                pre {{
+                    white-space: pre-wrap;
+                    word-wrap: break-word;
+                }}
+            </style>
+        <body>
+       <a href={}>out of band invitation url</a>
+        </body>
```
IngridPuppet, line 49: Please leave the URL plainly visible. It is more practical that way. See https://mediator.rootsid.cloud.

IngridPuppet (changes requested):
This PR changes the OOB URL into a format that is not normative. I left some comments regarding that and other points. Please, could you check?

crates/web-plugins/oob-messages/src/models.rs
```diff
@@ -112,7 +118,7 @@ where
 
     let did = diddoc.id.clone();
     let oob_message = OobMessage::new(&did);
-    let url: &String = &format!("{}:{}", server_public_domain, server_local_port);
+    let url: &String = &format!("{}", server_public_domain);
     let oob_url = OobMessage::serialize_oob_message(&oob_message, url)
```
Hermann-Core, line 122: The conversion here is unnecessary.
```suggestion
    let oob_url = OobMessage::serialize_oob_message(&oob_message, server_public_domain)
```

crates/web-plugins/oob-messages/src/models.rs
Hermann-Core, line 277:
```suggestion
        let oob_message: Vec<&str> = serialized_msg.split("/_oob").collect();
        let oob_message = oobmessage.get(1).unwrap().to_string();
```

crates/web-plugins/oob-messages/src/plugin.rs
```diff
@@ -17,12 +17,12 @@ pub struct OOBMessages {
 struct OOBMessagesEnv {
     storage_dirpath: String,
     server_public_domain: String,
-    server_local_port: String,
 }
 
 #[derive(Clone)]
 pub(crate) struct OOBMessagesState {
     pub(crate) filesystem: Arc<Mutex<dyn FileSystem>>,
+    pub(crate) oobmessage: String,
```
Hermann-Core, line 25: what is the purpose of the field you added here?

crates/web-plugins/oob-messages/src/web.rs
```diff
@@ -30,10 +36,14 @@ mod tests {
         std::env::set_var("STORAGE_DIRPATH", "tmp");
         std::env::set_var("SERVER_PUBLIC_DOMAIN", "example.com");
         std::env::set_var("SERVER_LOCAL_PORT", "8080");
-
-        let fs = MockFileSystem;
+        let server_public_domain = std::env::var("SERVER_PUBLIC_DOMAIN").unwrap();
+        let storage_dirpath = std::env::var("STORAGE_DIRPATH").unwrap();
+        let mut fs = MockFileSystem;
+        let oobmessage =
```
Hermann-Core, line 42:
```suggestion
        let oob_message =
```

Hermann-Core (changes requested):

crates/web-plugins/oob-messages/src/web.rs
Christiantyemele: I have check all the comments and from our discussions it seems this  PR actually solves no issue or brings and adjustments so, i guess it should be closed @Hermann-Core @IngridPuppet

IngridPuppet: I would agree. Also check the comment I left on the issue before planning to close it as well.

PR #290, opened 2025-01-13
Secure key storage

Christiantyemele:


IngridPuppet:
I had a look at one of the failing tests:

```
──── STDERR:             did-endpoint web::tests::verify_didpop
thread 'web::tests::verify_didpop' panicked at crates/web-plugins/did-endpoint/src/plugin.rs:39:50:
Secrets Mastet_KEY env variable required: NotPresent
```

I ran a troubleshooting locally and can confirm that it is indeed because a MASTER_KEY env variable is not fed to the test. I am not fully aware of your current setup for setting those variables during tests, but make sure you're also setting this one your changes expect.

The other failure message also appears related to your changes introducing a call to a `find_by_key` method. Do you have that properly mocked?

```
 ──── STDERR:             did-endpoint didgen::tests::test_did_validation
thread 'didgen::tests::test_did_validation' panicked at crates/web-plugins/did-endpoint/src/didgen.rs:267:5:
MockKeystore::find_key_by(Document({"kid": String("did:peer:123#key-1")}), [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]): No matching expectation found
```

Overall, I have another question: are all tests consistently passing on your local environments in the team, or you just expect them to pass on the pipeline? I'm asking because I feel something got broken in the meantime.

Christiantyemele:
> I had a look at one of the failing tests:
> 
> ```
> ──── STDERR:             did-endpoint web::tests::verify_didpop
> thread 'web::tests::verify_didpop' panicked at crates/web-plugins/did-endpoint/src/plugin.rs:39:50:
> Secrets Mastet_KEY env variable required: NotPresent
> ```
> 
> I ran a troubleshooting locally and can confirm that it is indeed because a MASTER_KEY env variable is not fed to the test. I am not fully aware of your current setup for setting those variables during tests, but make sure you're also setting this one your changes expect.
> 
> The other failure message also appears related to your changes introducing a call to a `find_by_key` method. Do you have that properly mocked?
> 
> ```
>  ──── STDERR:             did-endpoint didgen::tests::test_did_validation
> thread 'didgen::tests::test_did_validation' panicked at crates/web-plugins/did-endpoint/src/didgen.rs:267:5:
> MockKeystore::find_key_by(Document({"kid": String("did:peer:123#key-1")}), [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]): No matching expectation found
> ```
> 
> Overall, I have another question: are all tests consistently passing on your local environments in the team, or you just expect them to pass on the pipeline? I'm asking because I feel something got broken in the meantime.

the pipeline runs the tests with all the features enable while locally we run with the default features that's why `cargo test` might run locally and fail online because of some features

crates/keystore/src/lib.rs
```diff
@@ -127,6 +283,52 @@ pub mod tests {
...
+        }
+        async fn secure_store(
+            &self,
+            secrets: WrapSecret,
+            _master_key: [u8; 32],
+        ) -> Result<WrapSecret, RepositoryError> {
+            self.secrets.write().unwrap().push(secrets.clone().into());
+            Ok(secrets)
+        }
+    }
```
IngridPuppet, line 331: Condition mocks to a test build.
```rust
#[cfg(test)]
mod tests {
```

crates/web-plugins/did-endpoint/src/didgen.rs
```diff
@@ -79,8 +79,20 @@ where
...
 
     // Store authentication and agreement keys in the keystore.
-    store_key(auth_keys_jwk, &diddoc, &diddoc.authentication, keystore)?;
-    store_key(agreem_keys_jwk, &diddoc, &diddoc.key_agreement, keystore)?;
+    store_key(
+        auth_keys_jwk,
+        &diddoc,
+        &diddoc.authentication,
+        keystore,
+        master_key,
```
IngridPuppet, line 87: Don't you find it more appropriate to define the master_key as a property of the keystore? Once you have a keystore handle, it is not convenient to keep requiring its master key.

crates/web-plugins/did-endpoint/src/didgen.rs
```diff
@@ -130,7 +143,8 @@ where
 
     // Store the secret in the keystore
     task::block_in_place(move || {
-        Handle::current().block_on(async move { keystore.store(secret).await })
+        Handle::current()
+            .block_on(async move { keystore.secure_store(secret.into(), master_key).await })
     })
```
IngridPuppet, line 148: Most of the code relies on a KeyStore of Secrets. Changing that to a KeyStore of WrapSecret means you will keep converting between these formats outside of the KeyStore like here. Like I said before, I do not see the point in doing that.

crates/web-plugins/did-endpoint/src/didgen.rs
```diff
@@ -110,9 +122,10 @@ fn store_key<S>(
     diddoc: &Document,
     field: &Option<Vec<VerificationMethodType>>,
     keystore: &S,
+    master_key: [u8; 32],
 ) -> Result<(), Error>
 where
-    S: Repository<Secrets>,
+    S: SecureRepository<WrapSecret>,
 {
```
IngridPuppet, line 129: Expect a KeyStore, not a SecureRepository. At this point, how the KeyStore is implemented should not be constrained.

crates/web-plugins/did-endpoint/src/plugin.rs
```diff
@@ -36,6 +35,18 @@ fn get_env() -> Result<DidEndpointEnv, PluginError> {
         server_public_domain,
     })
 }
+pub fn get_master_key<'a>() -> Result<String, PluginError> {
+    let master_key = std::env::var("MASTER_KEY").expect("Secrets Mastet_KEY env variable required");
```
IngridPuppet, line 39: Bubble up the error instead of panicking. We're doing that for other env variables, right?
```suggestion
    let master_key = std::env::var("MASTER_KEY").expect("Secrets Master_KEY env variable required");
```

crates/web-plugins/did-endpoint/src/plugin.rs
```diff
@@ -44,11 +55,19 @@ impl Plugin for DidEndpoint {
 
     fn mount(&mut self) -> Result<(), PluginError> {
         let env = get_env()?;
+        let master_key = get_master_key()?;
+        let master_key = master_key.as_bytes().try_into().unwrap();
```
IngridPuppet, line 59: I suggest you directly return the master key as bytes from `get_master_key()`?

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
```diff
@@ -101,8 +102,13 @@ pub(crate) async fn process_mediate_request(
             kid: diddoc.key_agreement.first().unwrap().clone(),
             secret_material: agreem_keys_jwk,
         };
+        // Dummy master key
+        let master_key = get_master_key().unwrap().as_bytes().try_into().unwrap();
```
IngridPuppet, line 106: I think you don't want this comment anymore.

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
```diff
@@ -116,8 +122,13 @@ pub(crate) async fn process_mediate_request(
             kid: diddoc.authentication.first().unwrap().clone(),
             secret_material: auth_keys_jwk,
         };
+        // Dummy master key
+        let master_key = [0; 32];
```
IngridPuppet, line 126: ???

crates/web-plugins/didcomm-messaging/shared/src/utils/resolvers.rs
```diff
@@ -88,22 +87,28 @@ fn prepend_doc_id_to_vm_ids(diddoc: &mut Document) {
...
 #[async_trait]
 impl SecretsResolver for LocalSecretsResolver {
     async fn get_secret(&self, secret_id: &str) -> Result<Option<Secret>> {
+        // read master_key
+
+        let master_key = env::var("MASTER_KEY")
+            .unwrap()
+            .as_bytes()
+            .try_into()
+            .unwrap();
```
IngridPuppet, line 107: You do not want to do this every time a secret is resolved. Load that at the plugin level and pass that to the resolver constructor.

crates/web-plugins/did-endpoint/src/web.rs
```diff
@@ -100,17 +104,20 @@ async fn didpop(
 
         let kid = util::handle_vm_id(&method.id, &diddoc);
 
+        let master_key = get_master_key().unwrap().as_bytes().try_into().unwrap();
```
IngridPuppet, line 107: Prefer not loading this on every request. Consider keeping that in the state.

crates/web-plugins/didcomm-messaging/src/midlw.rs
```diff
@@ -188,13 +188,16 @@ pub async fn pack_response_message(
...
+
     use super::*;
     use shared::utils::tests_utils::tests::*;
 
     use serde_json::json;
 
     #[tokio::test]
     async fn test_pack_response_message_works() {
+        env::set_var("MASTER_KEY", "1234567890qwertyuiopasdfghjklxzc");
         let state = setup();
```
IngridPuppet, line 201: You are doing this again and again. Can't you set that env var inside the `setup()` method?

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tokio = { workspace = true, features = ["full"] }
 axum = { workspace = true, features = ["macros"] }
 uuid = { workspace = true, features = ["v4"] }
 hyper = { workspace = true, features = ["full"] }
+did-endpoint.workspace = true
```
IngridPuppet, line 27: Relocate the `get_master_key` method to avoid introducing this strange dependency. You may consider the shared utils crate for example.

storage/did.json
IngridPuppet, line 1: Do you mean to commit this file?

storage/oob_invitation.txt
IngridPuppet, line 1: Do you mean to commit this file?

storage/qrcode.txt
IngridPuppet, line 1: Do you mean to commit this file?

IngridPuppet (changes requested):

crates/database/src/lib.rs
```diff
@@ -28,6 +28,10 @@ pub enum RepositoryError {
     MissingIdentifier,
     #[error("target not found")]
     TargetNotFound,
+    #[error("decryption error")]
+    DecryptionError,
+    #[error("error deserializing to jwk")]
+    DeserializationError,
```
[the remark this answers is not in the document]
Christiantyemele: okay

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,172 @@
...
+    bson::{oid::ObjectId, to_vec, Bson, Document as BsonDocument},
+    Collection,
+};
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
-static SECRETS_COLLECTION: OnceCell<Collection<Secrets>> = OnceCell::new();
+static SECRETS_COLLECTION: OnceCell<Collection<WrapSecret>> = OnceCell::new();
```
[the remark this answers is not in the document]
Christiantyemele: This is what was before and @Hermann-Core  had a reason for doing this

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,172 @@
...
+
+/// Definition of a trait for secure repository operations.
+#[async_trait]
+pub trait SecureRepository<Entity>: Sync + Send
+where
+    Entity: Sized + Send + Sync + 'static + Clone,
```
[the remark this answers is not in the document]
Christiantyemele: Hum i don't understand what you mean @IngridPuppet

.github/scripts/test_config.sh
```diff
@@ -7,6 +7,7 @@ SERVER_LOCAL_PORT=3000
 STORAGE_DIRPATH=test/storage
 MONGO_URI=mongodb://localhost:27017/
 MONGO_DBN=didcomm-mediator
+MASTER_KEY=1234567890qwertyuiopasdfghjklxzc
 EOL
```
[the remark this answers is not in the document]
Christiantyemele: still passed as an env variable

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,172 @@
...
+        let collection = self.get_collection();
+
+        // Lock the Mutex and get the Collection
+        let collection = collection.read().await;
+
+        // Insert the new entity into the database
+        let metadata = collection.insert_one(secret_entity.clone()).await?;
+        if let Bson::ObjectId(oid) = metadata.inserted_id {
+            secret.set_id(oid);
+        }
```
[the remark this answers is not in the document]
Christiantyemele: :+1:

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,172 @@
...
+}
+impl Default for WrapSecret {
+    fn default() -> Self {
+        WrapSecret {
+            id: None,
+            kid: "".to_owned(),
+            secret_material: Vec::default(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
Christiantyemele: No actually it was for testing just

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,172 @@
...
+impl From<Secrets> for WrapSecret {
+    fn from(value: Secrets) -> Self {
+        let secret_material = to_vec(&value.secret_material).unwrap_or_default();
+        WrapSecret {
+            id: value.id,
+            kid: value.kid,
+            secret_material,
+        }
+    }
+}
```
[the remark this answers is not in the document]
Christiantyemele: yes it is encrypted, this conversion is used to transform from the secrets to wrapsecret which the secret_material will further be substituted with encrypted secrets in fact it is indistinghishable to know if the secrets in wrapSecrets are encrypted or not, but it is handled such that when we have a wrapsecret handle the secret are always encrypted

crates/keystore/src/lib.rs
```diff
@@ -42,23 +198,23 @@ where
...
-impl Default for KeyStore<Secrets> {
+impl Default for KeyStore<WrapSecret> {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl KeyStore<Secrets> {
+impl KeyStore<WrapSecret> {
     /// Create a new keystore with default Secrets type.
```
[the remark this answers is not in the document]
Christiantyemele: No you are mistaken what is moved around and what the application knows is wrapSecret(and this is also advantageous because one cannot distinghuish if its encrypted in not as it always looks encrypted) and all secrets impl are on wrapsecret only those who need secrets can have then with `.into` or from `find_key_by` function if they have access to which a private function

crates/keystore/src/lib.rs
```diff
@@ -42,23 +198,23 @@ where
...
         let collection = SECRETS_COLLECTION
             .get_or_init(|| {
                 let db = database::get_or_init_database();
                 let task = async move {
                     let db_lock = db.write().await;
-                    db_lock.collection::<Secrets>("secrets").clone()
+                    db_lock.collection::<WrapSecret>("secrets").clone()
                 };
                 tokio::task::block_in_place(|| tokio::runtime::Handle::current().block_on(task))
             })
```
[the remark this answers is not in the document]
Christiantyemele: This change is not from my issue i think you should discuss more of this with @Hermann-Core

crates/keystore/src/lib.rs
Christiantyemele: still not a change of mine

crates/web-plugins/did-endpoint/src/didgen.rs
Christiantyemele: It would have been best to have a global state for the master_key and all the other data which are needed but here for the master_key case there are some function which cannot have access to the keystore e.g the `get_secrets` function of `LocalDIDResolver` can not have access to keystore and i also find it dangerous

crates/web-plugins/did-endpoint/src/didgen.rs
Christiantyemele: `keyStore` is a struct which implements `SecureRepository`

crates/web-plugins/did-endpoint/src/didgen.rs
Christiantyemele: you're right but i can't do otherwise, except i will implement the functions on secrets instead of wrapsecret and that will still imply calls to `into`

crates/web-plugins/did-endpoint/src/plugin.rs
Christiantyemele: sure :+1:

crates/web-plugins/did-endpoint/src/plugin.rs
Christiantyemele: yeah it's better

crates/web-plugins/did-endpoint/src/web.rs
Christiantyemele: :+1: exactly

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
Christiantyemele: yes my bad

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/src/handler/stateful.rs
Christiantyemele: forgot this testing code

crates/web-plugins/didcomm-messaging/protocols/mediator-coordination/Cargo.toml
Christiantyemele: true we have to unchain the crates as much as possible

crates/web-plugins/didcomm-messaging/shared/src/utils/resolvers.rs
Christiantyemele: i could'nt say better

crates/web-plugins/didcomm-messaging/src/midlw.rs
Christiantyemele: i did'nt bother for that here as it's in the tests

storage/did.json
Christiantyemele: NO

IngridPuppet (approved):

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
 use async_trait::async_trait;
-use database::{Identifiable, Repository};
+use cocoon::MiniCocoon;
+use database::{Identifiable, Repository, RepositoryError};
 use did_utils::jwk::Jwk;
-use mongodb::{bson::oid::ObjectId, Collection};
+use mongodb::{
+    bson::{oid::ObjectId, to_vec, Bson, Document as BsonDocument},
```
Blindspot22, line 6: Please whats the use of `Bson`?
I don't see it been used anywhere

Blindspot22 (changes requested):

.github/scripts/test_config.sh
```diff
@@ -7,6 +7,7 @@ SERVER_LOCAL_PORT=3000
 STORAGE_DIRPATH=test/storage
 MONGO_URI=mongodb://localhost:27017/
 MONGO_DBN=didcomm-mediator
+MASTER_KEY=1234567890qwertyuiopasdfghjklxzc
```
Hermann-Core, line 10: The master key should not be passed as an env variable. This is a security risk.
Consider more secure approach such as [Docker secrets](https://docs.docker.com/engine/swarm/secrets/) or [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/).

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
...
+                secret_material: jwk,
+            };
+            // convert secret to entity
+            // let unwrap_entity: Entity = unwrap_secret.into();
+            Ok(Some(unwrap_secret))
+        } else {
+            Ok(None)
+        }
+    }
+}
```
Hermann-Core, line 96: What is the need of defining another trait with the same bounds for the secrets handling?
I see a lot of unwrap in your code. Please handle the errors gracefully. **Don't panic!**

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
...
+where
+    T: Sized + Clone + Send + Sync + 'static,
+    T: Identifiable + Unpin + From<Secrets>,
+    T: From<WrapSecret> + Into<WrapSecret> + Into<Secrets>,
+    T: Serialize + for<'de> Deserialize<'de>,
+{
+    fn get_collection(&self) -> Arc<RwLock<Collection<T>>> {
+        Arc::new(RwLock::new(self.collection.clone()))
+    }
+}
```
Hermann-Core, line 108: I see a lot a duplicated code. I really think that this can be optimized.

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
...
+        let secret_material: BsonDocument = bson::from_slice(&secret_material).unwrap();
+        let secret_material = serde_json::to_value(secret_material).unwrap();
+        let secret_material: Jwk = serde_json::from_value(secret_material).unwrap();
+        Secrets {
+            id: value.id,
+            kid: value.kid,
+            secret_material,
+        }
+    }
+}
```
Hermann-Core, line 132: Conversions that can fail should use the **`TryFrom`** trait instead to take into account the errors that may be encountered during the conversion.

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
...
+impl Identifiable for WrapSecret {
+    fn id(&self) -> Option<ObjectId> {
+        self.id
+    }
+
+    fn set_id(&mut self, id: ObjectId) {
+        self.id = Some(id);
+    }
+}
```
Hermann-Core, line 154: Too many duplicated code. Pretty sure we can avoid this.

crates/web-plugins/did-endpoint/src/didgen.rs
```diff
@@ -110,9 +122,10 @@ fn store_key<S>(
     diddoc: &Document,
     field: &Option<Vec<VerificationMethodType>>,
     keystore: &S,
+    master_key: [u8; 32],
```
Hermann-Core, line 125: You should avoid passing the master key everywhere like this. It should only be used internally to decrypt the secrets before returning them to the caller. It should not remain in memory for long to mitigate security risks.

crates/keystore/src/lib.rs
```diff
@@ -1,16 +1,160 @@
...
+    async fn secure_store(
+        &self,
+        entity: Entity,
+        master_key: [u8; 32],
+    ) -> Result<Entity, RepositoryError> {
+        // convert and entity into a Wrapsecret
+        let secret: WrapSecret = entity.into();
+
+        // hardecoded seed to always correspond to master key
+        let seed = [0; 32];
```
Hermann-Core, line 36: Not a good idea at all to hardcode a seed like this. Generate a random seed instead.

Hermann-Core (changes requested):

crates/keystore/src/lib.rs
```diff
@@ -42,23 +186,23 @@ where
     collection: Collection<T>,
 }
 
-impl Default for KeyStore<Secrets> {
+impl Default for KeyStore<WrapSecret> {
```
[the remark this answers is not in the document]
Christiantyemele: The secret we had before were in plaintext, which is an issue so i had to convert the secrets into a more protected type `wrapsecret` when which is what the application knows only the functions needed `secret` can the decrypt and decode the wrapsecret, this is avoid exposing the secrets and for storing them protected

crates/keystore/src/lib.rs
Christiantyemele: this is something which actually needs to be removed

crates/keystore/src/lib.rs
Christiantyemele: i though the errors were fatal but thinking of it again i guest there are not. i did not want to implement this methods in the existing crate so as to keep things seperate as that would will require me bringing the keystore into the database which is what we are runing from

.github/scripts/test_config.sh
Hermann-Core: No. Sensitive data should not be passed as env variables.

crates/keystore/src/lib.rs
Hermann-Core: At this level, all the errors should be handled as the errors that can happen here are logic errors and should be catch-ed and backtrack-ed.

crates/keystore/src/lib.rs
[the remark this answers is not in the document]
Blindspot22: Kk is I see

crates/keystore/src/lib.rs
Blindspot22: Kk. You can remove it then

Hermann-Core:
Currently, the keystore does not use any encryption backend, but I have implemented the AWS KMS encryption system. Once I will get the credentials, we should be able to easily switch to it.

.github/scripts/test_config.sh
Christiantyemele: Then this should be removed

crates/keystore/Cargo.toml
```diff
@@ -11,8 +11,11 @@ async-trait.workspace = true
 mongodb.workspace = true
 database.workspace = true
 serde.workspace = true
+eyre.workspace = true
+thiserror.workspace = true
 serde_json.workspace = true
+aws-sdk-kms.workspace = true
```
Christiantyemele, line 17: This requires rust version 1.81.1 or higher please also modify this in the Dockerfile

crates/keystore/src/encryptor/plaintext.rs
```diff
@@ -0,0 +1,16 @@
+use async_trait::async_trait;
+
+use crate::{Error, KeyEncryption};
+
+pub(crate) struct NoEncryption;
```
Christiantyemele, line 5: Can you say what this NoEncryption struct is for ?

crates/keystore/src/repository/no_repo.rs
```diff
@@ -0,0 +1,22 @@
+use async_trait::async_trait;
+
+use crate::Error;
+
+use super::SecretRepository;
+
+pub(crate) struct NoRepository;
```
Christiantyemele, line 7: what is this NoRepository struct for ?

Hermann-Core: It is self descriptive I think.

crates/keystore/src/encryptor/plaintext.rs
Hermann-Core: It is self descriptive though.

crates/keystore/src/repository/no_repo.rs
Christiantyemele: > It is self descriptive I think.

Not enough cause  i do not know why we need a NoEncryption and and Encryptions struct, what does the NoEncryption struct stores ?

crates/keystore/Cargo.toml
Christiantyemele: @Hermann-Core you have'nt addresses this

Christiantyemele: @Hermann-Core actions are failing are you aware ?

Hermann-Core: Reverse the last commits you did. They are the cause

Christiantyemele: > Reverse the last commits you did. They are the cause

what commits ?, i think you should resolve them since you are the one handling the issue now,

Christiantyemele: > Reverse the last commits you did. They are the cause

i have reverted them

Hermann-Core:
Closed by #329

Hermann-Core:
Closed by #329

PR #295, opened 2025-01-16
docs(): added architecture doc

Christiantyemele:


docs/architecture.md
IngridPuppet, line 1: The issue seems to be concerned about a "high-level overview of [the mediator's] crates and explaining the function of each crate." How is that achieved by this writeup? You seem focused on the hierarchical plugin architecture instead.

docs/architecture.md
```diff
@@ -0,0 +1,65 @@
+
+# **Architecture Documentation: Plugin-Based DIDComm Router**
+
+## **1. Introduction**
+### **1.1 Purpose**
```
IngridPuppet, line 5: Are these star symbols necessary? Titles should be bold by default.

docs/architecture.md
Christiantyemele: yes i think the title is misleading then and should be addressed

IngridPuppet (approved):

PR #304, opened 2025-02-03
fix():removing un used code and fixing warnings

ndefokou:
@Hermann-Core

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -47,7 +47,11 @@ jobs:
       run: cargo fmt --all -- --check
 
     - name: Build Project
-      run: cargo build --workspace --all-features
+      run: cargo build --workspace --all-features -D warning
 
     - name: Run Tests
       run: cargo nextest run --workspace --all-features
+    
+    -name: check for warnings in code 
+      run: |
+        cargo clippy --workspace --all-targets --all-features -- -D warnings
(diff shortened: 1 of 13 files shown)


.github/workflows/CI.yml
```diff
@@ -47,7 +47,11 @@ jobs:
       run: cargo fmt --all -- --check
 
     - name: Build Project
-      run: cargo build --workspace --all-features
+      run: cargo build --workspace --all-features -D warning
```
Hermann-Core, line 50: Is this command working as expected? I could not reproduce it locally.
The right approach here is to define a RUSTFLAGS env variable and add **-D warnings** as flag as suggested [here](https://github.com/rust-lang/cargo/issues/8424).

.github/workflows/CI.yml
```diff
@@ -47,7 +47,11 @@ jobs:
...
 
     - name: Run Tests
       run: cargo nextest run --workspace --all-features
+    
+    -name: check for warnings in code
```
Hermann-Core, line 55:
```suggestion
    - name: Lints checks
```
Don't forget to add space between the dash and **name**.

Hermann-Core (changes requested):

Hermann-Core:
Also consider adding a step for `cargo doc` also to ensure that the API documentation is functional.

.github/workflows/CI.yml
ndefokou: Yes it is a good idea

Hermann-Core (approved):

PR #311, opened 2025-02-04
Remove unnecessary memory synchronization primitives from mongodb types

Hermann-Core:


IngridPuppet (approved):

crates/database/src/lib.rs
```diff
@@ -30,20 +28,20 @@ pub enum RepositoryError {
     TargetNotFound,
 }
 
-static MONGO_DB: OnceCell<Arc<RwLock<Database>>> = OnceCell::new();
+static MONGO_DB: OnceCell<Database> = OnceCell::new();
```
Christiantyemele, line 32: why remove the RwLock

Ngha-Boris (approved):

crates/database/src/lib.rs
Hermann-Core: > why remove the RwLock

the `Database` type is already thread-safe and designed for concurrent use. Wrapping it with the `RwLock` adds 
unnecessary overhead.

Ogenbertrand (approved):

PR #329, opened 2025-02-17
Implement a secure key storage

Hermann-Core:


crates/keystore/src/repository/mongodb.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// The secret collection will be initialized once even if this function is called multiple times.
+    pub(crate) fn new() -> Self {
+        static SECRETS_COLLECTION: OnceCell<Collection<Secrets>> = OnceCell::new();
+        let db = database::get_or_init_database();
+        let collection = SECRETS_COLLECTION
+            .get_or_init(|| {
+                let task = async move { db.collection::<Secrets>("secrets").clone() };
+                block_in_place(|| Handle::current().block_on(task))
+            })
+            .clone();
```
IngridPuppet, line 29: I believe this is overengineering. How heavy do you think linking to the collection in the database is to warrant these operations? 
```rust
db.collection::<Secrets>("secrets")
```
Also, can't you just inject the database as an argument instead of relying on the singleton pattern?

crates/keystore/src/tests.rs
```diff
@@ -0,0 +1,120 @@
...
+        let index = secrets.iter().position(|(k, _)| k == kid);
+        match index {
+            Some(index) => {
+                secrets.remove(index);
+                Ok(())
+            }
+            None => Err(Error::msg(
+                ErrorKind::RepositoryFailure,
+                format!("Secret with kid {kid} not found"),
+            )),
```
IngridPuppet, line 54: Is this behavior matching the concrete repository impl? I think that was removed and not finding the resource not considered an error.

crates/keystore/src/tests.rs
```diff
@@ -0,0 +1,120 @@
...
+            "x": "Z0GqpN71rMcnAkky6_J6Bfknr8B-TBsekG3qdI0EQX4",
+            "d": "fI1u4riKKd99eox08GlThknq-vEJXcKBI28aiUqArLo"
+        }))
+        .unwrap()
+    }
+
+    #[tokio::test]
+    async fn test_keystore_flow() {
+        let jwk1 = secret1();
+        let jwk2 = secret2();
```
IngridPuppet, line 89: I'm surprised you aren't also testing that keys are not persisted in clear. I could not find tests at the encryptor module layer either.

crates/web-plugins/didcomm-messaging/shared/src/breaker.rs
```diff
@@ -34,15 +34,15 @@ use tokio::time::Sleep;
...
+/// *   [`CircuitBreaker::half_open_max_failures(self, max_retries: usize)`]: Sets the maximum number of attempts in half-open state
 ///     before reopening the circuit.
-/// *   [`reset_timeout(self, reset_timeout: Duration)`](timeout): Sets the duration the circuit remains open after tripping.
+/// *   [`CircuitBreaker::reset_timeout(self, reset_timeout: Duration)`]: Sets the duration the circuit remains open after tripping.
 ///     After this timeout, the circuit transitions to the half-open state.
-/// *   [`exponential_backoff(self, initial_delay: Duration)`](exponential_backoff): Configures an exponential backoff strategy for retries.
+/// *   [`CircuitBreaker::exponential_backoff(self, initial_delay: Duration)`]: Configures an exponential backoff strategy for retries.
 ///     The delay between retries increases exponentially. This overrides any previously set backoff.
-/// *   [`constant_backoff(self, delay: Duration)`](constant_backoff): Configures a constant backoff strategy for retries.
+/// *   [`CircuitBreaker::constant_backoff(self, delay: Duration)`]: Configures a constant backoff strategy for retries.
```
IngridPuppet, line 45: Please refrain from adding changes not related to your PR. It leaves the reviewer confused, especially when they are not qualified to review them.

src/main.rs
```diff
@@ -71,42 +71,3 @@ fn config_tracing() {
...
-mod test {
-
-    use reqwest::Client;
-    use tokio::{task, time::Instant};
-
-    #[tokio::test]
-    async fn test() {
-        let client = Client::new();
-        let url = "https://didcomm-mediator.eudi-adorsys.com/";
-        let num_requests = 1000;
```
IngridPuppet, line 85: I understand the test is problematic, but I do not think this is the PR for removing it.

.gitignore
```diff
@@ -33,6 +33,7 @@ test/
 
 # Environment variables files 
 .env
+.env.example
```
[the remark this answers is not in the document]
Hermann-Core: I don't think that we require it anymore since it is constructed dynamically in the pipeline for the tests.

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.4
 
 # Stage 1: Build
-FROM rust:1.80-alpine AS builder
+FROM rust:1.83-alpine AS builder
```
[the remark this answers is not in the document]
Hermann-Core: One dependency that I used required a higher Rust version. So I've updated the builder image version to follow that requirement.

crates/database/src/lib.rs
```diff
@@ -26,6 +26,10 @@ pub enum RepositoryError {
     MissingIdentifier,
     #[error("target not found")]
     TargetNotFound,
+    #[error("decryption error")]
+    DecryptionError,
+    #[error("error deserializing to jwk")]
+    JwkDeserializationError,
```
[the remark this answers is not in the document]
Hermann-Core: I don't get it.

crates/keystore/src/lib.rs
```diff
@@ -1,241 +1,151 @@
...
+    where
+        T: Serialize + DeserializeOwned + Send + Sync + 'static,
+    {
+        let mock_repository = MockSecretRepository::new(secrets);
+        let encryptor = NoEncryption;
+        Self {
+            repository: Arc::new(mock_repository),
+            encryptor: Arc::new(encryptor),
+        }
     }
```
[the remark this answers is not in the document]
Hermann-Core: Makes sense.

crates/keystore/src/repository/mongodb.rs
Hermann-Core: This is to ensure first that we initialize the secret collection only once, also ensuring that we have only one secrets collection.

crates/keystore/src/tests.rs
```diff
@@ -0,0 +1,120 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use std::sync::RwLock;
+
+#[derive(Default)]
+pub struct MockSecretRepository<T>
+where
+    T: Serialize + DeserializeOwned,
+{
+    secrets: RwLock<Vec<(String, T)>>,
+}
```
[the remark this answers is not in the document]
Hermann-Core: I want interior mutability on the vector since the methods of the `SecretRepository` trait take `self` by reference. I could have used something cheaper like `RefCell` but the tests don't run for long so I think it is fine.

crates/keystore/src/tests.rs
Hermann-Core: Just checked the delete function of the Repository and you are right. Let me fix that.
But anyways it is just for the test. In real scenarios, it will not throw an error.

crates/keystore/src/tests.rs
Hermann-Core: Hmm right. I only wanted to test the keystore flow. But I can also provide a mock encryptor.

crates/web-plugins/didcomm-messaging/shared/src/breaker.rs
Hermann-Core: We added a step in the CI to check the docs too. These changes have been done to make the pipeline happy.

src/main.rs
Hermann-Core: I think it appears when I fetched the main.

.gitignore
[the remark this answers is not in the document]
IngridPuppet: It serves a guide for users to create a `.env` in the local environment. It is a mistake that any code is relying on it.

crates/database/src/lib.rs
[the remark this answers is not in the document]
IngridPuppet: Christian added them in his first implementation. I don't think they are used anymore.

IngridPuppet (commented):
@Hermann-Core - Just reviewed again. Could you check the conversations I left unresolved?

.gitignore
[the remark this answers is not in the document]
Hermann-Core: > It serves a guide for users to create a `.env` in their local environment. It is a mistake that any code is relying on it.

Don't you think that it is better to add this in the **README** description instead? Along with all the required variables. This is how many projects I've se do.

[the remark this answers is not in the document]
IngridPuppet: It is actually very common to have a `.env.example` or `.env.dist` file committed for that purpose. If you make the information available in README, then I'm okay with that.

crates/database/src/lib.rs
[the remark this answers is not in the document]
Hermann-Core: Not really. I’ve reimplemented it completely. But still. Let me remove it.

IngridPuppet (approved):

README.md
```diff
@@ -55,16 +53,53 @@ git clone git@github.com:adorsys/didcomm-mediator-rs.git
...
+Another option is to use [Docker](https://www.docker.com):
+
+```sh
+docker pull mongo
+docker run --name mongodb -d mongo
+```
+
+### Environmental variables
+
+You need to create a **`.env`** file in the root directory of the project and add the following variables:
```
Christiantyemele, line 73: Please only keep the changes related to this issue.

Christiantyemele (changes requested):

README.md
Hermann-Core: Of course they are. I modified the `.env` and `.env.example` files and removed them from the Git index. Since this change affects how the server can be started, I had to inform users about the importance of the `.env` file by updating the README. See my comments with Ingrid above for more details.

crates/keystore/src/repository.rs
```diff
@@ -8,7 +8,12 @@ use crate::Error;
 /// Abstract interface for secret storage backends.
 #[async_trait]
 pub trait SecretRepository: Send + Sync {
+    /// Store a given secret bytes in the repository.
```
IngridPuppet, line 11:
```suggestion
    /// Store a given secret's bytes in the repository.
```

crates/keystore/src/repository.rs
```diff
@@ -8,7 +8,12 @@ use crate::Error;
...
     async fn store(&self, kid: &str, key: &[u8]) -> Result<(), Error>;
+
+    /// Retrieve a secret bytes from the repository.
```
IngridPuppet, line 14:
```suggestion
    /// Retrieve a secret's bytes from the repository.
```

IngridPuppet (approved):

IngridPuppet (approved):

Christiantyemele (approved):

PR #344, opened 2025-03-06
Fix/issue 332

Ogenbertrand:
We get this error "message must not be anonencypt" during the forward protocol, because the unpacking middle way expect massages to be authcrpt but some protocols like the forward protocol are anoncrypt, so we need to modify the check below to handle that case

### DIFF
--- a/crates/web-plugins/didcomm-messaging/src/midlw.rs
+++ b/crates/web-plugins/didcomm-messaging/src/midlw.rs
@@ -135,7 +135,6 @@ async fn unpack_payload(
         return Err(response.into_response());
     }
 
-    const ROUTING_PROTOCOL_MSG_TYPE: &str = "https://didcomm.org/routing/2.0/forward";
     if plain_message.type_ != ROUTING_PROTOCOL_MSG_TYPE
         && (plain_message.from.is_none() || !metadata.authenticated || metadata.anonymous_sender)
     {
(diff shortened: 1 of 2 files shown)


crates/web-plugins/didcomm-messaging/src/midlw.rs
```diff
@@ -133,7 +133,9 @@ async fn unpack_payload(
         return Err(response.into_response());
     }
 
-    if plain_message.from.is_none() || !metadata.authenticated || metadata.anonymous_sender {
+    if plain_message.type_ != "https://didcomm.org/routing/2.0/forward"
```
Hermann-Core, line 136: I suggest that you define that string literal as a `const` for better readability.
```suggestion
    if plain_message.type_ != ROUTING_2_0_MSG_TYPE
```

Hermann-Core (changes requested):

crates/web-plugins/didcomm-messaging/src/midlw.rs
```diff
@@ -133,7 +135,10 @@ async fn unpack_payload(
         return Err(response.into_response());
     }
 
-    if plain_message.from.is_none() || !metadata.authenticated || metadata.anonymous_sender {
+    const ROUTING_PROTOCOL_MSG_TYPE: &str = "https://didcomm.org/routing/2.0/forward";
```
Hermann-Core: What this is still doing here? Carefully check what you do. That’s why the pipeline is failing.

Hermann-Core (changes requested):

Hermann-Core (approved):

PR #347, opened 2025-03-17
feat: release workflow

Blindspot22:


.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      uses: actions/checkout@v3
+
+    - name: Set up Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+
+    - name: Install Dependencies
+      run: cargo build --release
```
Hermann-Core, line 30: The name does not align with what is being run.

Hermann-Core (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+on:
+  schedule:
+    # Runs every 3 weeks on Monday at 12:00 UTC
+    - cron: "0 12 * * 1"
+
+  workflow_dispatch: # Allows manual triggering of the workflow
+    inputs:
+      release_notes:
+        description: 'This is the initial release of didcomm-mediator-rs, a Rust implementation of a mediator for the DIDComm v2 protocol. The mediator facilitates secure, decentralized communication by managing routing and keylist functionality for agents in a Self-Sovereign Identity (SSI) ecosystem.'
```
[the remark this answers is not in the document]
Blindspot22: Okay :+1:

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Set up Rust
+      uses: actions-rs/toolchain@v1
```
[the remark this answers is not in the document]
Blindspot22: Okay

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+        required: false
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
```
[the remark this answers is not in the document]
Blindspot22: Yeah sure 
https://github.com/actions/checkout

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
+name: Automatic Release
+
+on:
+  schedule:
+    # Runs every 3 weeks on Monday at 12:00 UTC
+    - cron: "0 12 * * 1"
```
Awambeng, line 6: Please try correcting the cron job to run as expected.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Build the project
+      run: cargo build --release
+
+    - name: Bump Version
+      id: version
+      run: |
+        version=$(cargo pkgid | grep -oP '(?<=#).*') # Extract current version
+        new_version=$(echo $version | awk -F. '{$NF+=1; print $0}' OFS=.) # Increment patch
+        echo "new_version=$new_version" >> $GITHUB_ENV
+        echo "::set-output name=version::$new_version"
```
Awambeng, line 38: Don't you think this is a repetition? Also the `::set-output` action is deprecated so i will suggest we go with the first option. WDYT?

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
+name: Automatic Release
```
Awambeng, line 1:
```suggestion
name: Release Workflow
```

.github/workflows/release.yml
Awambeng, line 38: Please try formatting your code

Awambeng, line 38: Is there not already existing actions for versioning?

Awambeng (changes requested):
Thanks for your PR @Blindspot22. Overall, it looks good. I’ve added some comments—please take a look when you get a chance

Also, please ensure everything works as expected by testing before merging. Let me know if you need any help!

.github/scripts/cron_script.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Check if the current week is divisible by 3
+if (( $(date +%U) % 3 == 0 )); then
+  # Your actual task here
+  echo "Running the job as it's the 3rd week."
+else
+  echo "Skipping this week."
+fi
```
Awambeng, line 9: This is too complex why not just use:
```
on:
  schedule:
    - cron: "0 12 * * 1#3"
```
Reference for using cron jobs: https://crontab.cronhub.io/

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+        exit 0
+
+    - name: Build the project
+      run: cargo build --release
+
+    - name: Run Cron Script
+      run: |
+        set -e # Exit on any error
+        chmod +x ./scripts/cron_script.sh
+        ./scripts/cron_script.sh
```
Awambeng, line 61: This is obsolete

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+        override: true
+
+    - name: Check for Changes
+      id: check_changes
+      run: |
+        # Fetch tags to compare with the latest release
+        git fetch --tags
+        
+        # Get the latest tag and compare it with the current branch
+        latest_tag=$(git describe --tags $(git rev-list --tags --max-count=1))
```
Awambeng, line 36: how will this behave if the repository has no tag yet?

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+  
+          // Update version in Cargo.toml
+          const fs = require("fs");
+          const cargoToml = fs.readFileSync("Cargo.toml", "utf-8");
+          const updatedCargoToml = cargoToml.replace(/version = ".*"/, `version = "${newVersion}"`);
+          fs.writeFileSync("Cargo.toml", updatedCargoToml);
+  
+          // Commit and push changes
+          await exec.exec("git", ["commit", "-am", `Release ${newVersion}`]);
+          await exec.exec("git", ["push", "origin", `release/${newVersion}`]);
```
Awambeng, line 93: I am not too comfortable with this but if it works then fine.
You can get more inspiration from this work https://rapidrecast.io/blog/simplify-rust-releases-with-github-actions/

Awambeng (dismissed):

.github/scripts/cron_script.sh
Blindspot22: Ohhh okay thanks

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+            target            
+          key: ${{ runner.os }}-${{ runner.arch }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+          restore-keys: |
+            cargo-build-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/Cargo.toml') }}
+            cargo-build-${{ runner.os }}-${{ runner.arch }}-            
+
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
Awambeng, line 176: this is no longer maintained can you switch back to the action you were using

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+    needs: test
+    runs-on: ubuntu-latest
+    outputs:
+      rr_cargo_version: ${{ steps.get-version.outputs.VERSION }}
+      workflow_git_tag: ${{ steps.get-version.outputs.WORKFLOW_GIT_TAG }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.RELEASE_TOKEN }}
```
Awambeng, line 73: Does this secret exist? why don't you use the GITHUB_TOKEN secret ?

Awambeng (changes requested):

.github/workflows/release.yml
Blindspot22: okay

.github/workflows/release.yml
Blindspot22: Yes it does!
I created one just for this project

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type of release (major/minor/patch)'
+        required: true
+        default: 'minor'
+      dry_run:
+        description: 'Dry run (true/false)'
+        required: true
+        default: 'true'
```
Awambeng, line 16: You define a dry_run input but it's currently unused in the workflow logic. Can i know why this was defined?

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+on:
+  push:
+    tags:
+      - 'v*'  # Trigger on tag push like v0.1.0
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'Type of release (major/minor/patch)'
+        required: true
+        default: 'minor'
```
Awambeng, line 12: This type input is currently not used in the workflow. If it's not tied to version bumping logic (e.g., using cargo-release or semantic versioning tools), consider removing it to keep the workflow clean.

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Run Tests
+        if: ${{ github.event.inputs.skip_tests != 'true' }}
+        run: cargo test --verbose
+
+      - name: Save Release Binary
+        if: startsWith(github.ref, 'refs/tags/v')
+        run: |
+          mkdir -p release-artifacts
+          cp target/release/didcomm-mediator release-artifacts/
```
Awambeng, line 61: You may want to ensure the binary has the correct executable permissions before uploading
```suggestion
      - name: Save Release Binary
        if: startsWith(github.ref, 'refs/tags/v')
        run: |
          mkdir -p release-artifacts
          cp target/release/didcomm-mediator release-artifacts/
          chmod +x release-artifacts/didcomm-mediator
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+  create_release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    needs: build_and_test
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - name: Download Binary Artifact
+        uses: actions/download-artifact@v4.1.8
+        with:
+          name: didcomm-mediator-binary
```
Awambeng, line 79: Since this workflow was aimed at supporting multiple platforms it will be good if we can differentiate their names 
(e.g., didcomm-mediator-linux-x64) for linux distro

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+          toolchain: stable
+
+      - name: Cache Cargo
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/index
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
Awambeng, line 48: Consider including the Rust version in the cache key to avoid using stale builds when the toolchain changes. You could do this by extracting the version with `rustc --version` and passing it into the cache key.

Awambeng (dismissed):

.github/workflows/release.yml
Blindspot22: The dry_run input was included for future enhancements to allow testing the workflow without making changes to the repository or publishing artifacts. It's currently unused but will be integrated in subsequent updates.

But if you wish, I could remove it for now.

.github/workflows/release.yml
Blindspot22: Okay on it

.github/workflows/release.yml
Blindspot22: Okay on it

.github/workflows/release.yml
Blindspot22: For now there's no specific platform support, Just the artifacts.

.github/workflows/release.yml
Blindspot22: Adding the Rust version to the cache key introduces more complexity to the workflow. We would also need to add a step to retrieve the version (rustc --version) and pass it dynamically to the cache key.

I would suggest that, since we're consistently using the stable toolchain and rarely upgrade, the current setup balances simplicity and efficiency. We can revisit this if we start encountering issues with stale caches in the future.

.github/workflows/release.yml
Blindspot22: Done

.github/workflows/release.yml
Blindspot22: Done

PR #383, opened 2025-05-15
feat: make server address configurable via SERVER_HOST env var

Blindspot22:


Blindspot22:
Done in another ticket under (Status-List Server)

Issue #380, opened 2025-05-06
Make server address configurable


Hermann-Core:
Currently, the server's address appears to be hardcoded in the application code, rather than being loaded from some configuration mechanism. This limits the flexibility of the server configuration depending on the target environment for example.

PR #384, opened 2025-05-15
feat: make server address configurable via SERVER_HOST env var

Blindspot22:


Hermann-Core:
It looks like you're working in the wrong repository. This have to be implemented in the **status list server** repository instead.

Blindspot22:
> It looks like you're working in the wrong repository. This have to be implemented in the **status list server** repository instead.

Well the ticket was created under the mediator-repository rather
See here https://github.com/adorsys/didcomm-mediator-rs/issues/380

Blindspot22:
> It looks like you're working in the wrong repository. This have to be implemented in the **status list server** repository instead.

Should we do a transfer?
Like we close this one and create a new ticket with the same demand under the **status  list server** repository?

Hermann-Core:
> Should we do a transfer? Like we close this one and create a new ticket with the same demand under the **status list server** repository?

Yes

PR #398, opened 2025-06-17
393 update keystore to use aws secrets manager

Hermann-Core:
I added the support for AWS Secrets Manager service and caching support for the mediator secrets. Currently the sercrets are stored in plaintext in the database.

### DIFF
--- a/crates/keystore/src/repository/aws.rs
+++ b/crates/keystore/src/repository/aws.rs
@@ -1,0 +1,93 @@
+use std::{num::NonZeroUsize, time::Duration};
+
+use async_trait::async_trait;
+use aws_config::SdkConfig;
+use aws_sdk_secretsmanager::{
+    operation::get_secret_value::GetSecretValueError, Client as SecretsClient,
+    Config as SecretsConfig,
+};
+use aws_secretsmanager_caching::SecretsManagerCachingClient as SecretsCacheClient;
+use eyre::eyre;
+use tracing::warn;
+
+use crate::{Error, ErrorKind};
+
+use super::SecretRepository;
+
+/// Type used for AWS Secrets Manager operations
+pub(crate) struct AwsSecretsManager {
+    client: SecretsClient,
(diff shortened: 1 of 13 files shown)


hugoib:
My understanding of this MR is that will make local running versions unusable, unless we share files with keys and do some tricky config that should be more straightforward. 

Meaning, if I want to run it locally, what is the easiest way to achieve this  (and for further users / documented). Can your solution also provide and easy way to run locally for testing purposes?

crates/keystore/src/repository/aws.rs
```diff
@@ -0,0 +1,93 @@
...
+                    .create_secret()
+                    .name(name)
+                    .secret_string(secret)
+                    .send()
+                    .await?;
+                Ok(())
+            }
+            Err(sdk_err) => Err(sdk_err.into()),
+        }
+    }
```
Ngha-Boris, line 65: Currently, the delete method refreshes the cache after storing a secret, meaning cached values might be stale until the TTL expires.
I suggest that for consistency, consider refreshing the cache after a successful `put_secret_value` call in the store methos, similar to the delete method.

Ngha-Boris: If possible, can you please ensure tests cover the following scenarios:
   1.  Storing a new key and retrieving it.
   2.  Retrieving a non-existent key (should return None).
   3.  Deleting a key and verifying it’s no longer retrievable.

Hermann-Core: > If possible, can you please ensure tests cover the following scenarios:
> 
> 1. Storing a new key and retrieving it.
> 2. Retrieving a non-existent key (should return None).
> 3. Deleting a key and verifying it’s no longer retrievable.

Rest assured that all these test cases have been covered. While these could not be unit tested directly, I opted to validate them through integration tests using an AWS services emulation tool to ensure realistic behavior and interaction with the AWS APIs.

PR #402, opened 2025-06-20
399 add a centralized storage for the mediator did document

### LINKED ISSUE #399: Add a centralized storage for the mediator DID Document
We should avoid storing the mediator `DID Document` directly in the filesystem as each instance of the mediator will have its own DID Document which is not practical.

Hermann-Core:


ndefokou (commented):
Wow All good on my side Mr @Hermann-Core But i think you should resolve the conflicts you are having before we can merge

ndefokou (approved):

Hermann-Core:
> 1. Do we need migration logic? (from previous approach to new approach). My take is no since we don't have a lot of deployments but it is worth to ask and confirm.
> 2. we no longer need STORAGE_DIRPATH, right?
> 3. I wonder if the change of port (3000 / 8080) (why are we changing it?) in the docker-compose is going to break something in our deployment
> 
> I reviewed the approach and codebase, now I want to run it locally. I might have issues... asking for support.
> 
> Screenshot 2025-07-03 at 10 40 10
> Will report back once I can test functionality.

Thanks for your comment, I’ll try to address your questions:

1. We need to update the deployment to eliminate the dependency on the local filesystem. That includes:
   - Ensuring AWS credentials are passed to enable Secrets Manager usage.
   - Removing the `STORAGE_DIRPATH` environment variable from the deployment configurations, as it's no longer needed.

2. Yes we don't need it anymore.

3. The deployment via Helm is independent from the Docker Compose setup. However, the Compose configuration was updated because:
   - The server binds to port `3000` as defined in the `Dockerfile`.
   - Compose should reflect that, so we can run the app locally without confusion.

### For Local Testing

The updated `docker-compose.yml` now includes all necessary services for running the mediator locally. To get started:

- Create a `.env` file in the project root.
- Copy the content from `.env.example`, which includes safe defaults.
- Run:

```sh
docker compose up --build
```
Let me know if you have successfully run it locally

hugoib:
I ran into issues trying to run it, I reached out to Hermann for support.

hugoib (approved):

hugoib:
Approved, I tested the functionality and the did document. I am not up to date with the pipeline of this project and the deployment, I don't know if it will break something or what needs to be adapted in terms of deployment given this change.

If there are no dealbreakers on that side, it can be merged.

ndefokou (approved):

